PR #5, opened 2016-10-20
Pylint cli package

dankilman:


aria/cli/commands.py
```diff
@@ -210,7 +226,7 @@ def __call__(self, args_namespace):
                     args_namespace.workflow_id,
                     deployment.workflows.keys()))
 
-        workflow_parameters = self._merge_and_validate_execution_parameters(
+        workflow_parameters = self._merge_and_validate_parameters(
```
dankilman: restore name

aria/cli/config.py
```diff
@@ -32,6 +37,11 @@
 
 
 def load_configurations():
+    """
+    Dynamically load attributes into the config module from the ``config.yaml`` defined in the user
+    configuration directory
+    :return:
```
dankilman: remove return

PR #7, opened 2016-10-20
pylint aria.storage

mxmrlv:


aria/storage/__init__.py
```diff
@@ -99,12 +102,16 @@ def __getattr__(self, table):
         return super(ModelStorage, self).__getattr__(table)
 
     def register(self, model_cls):
-        # todo: add documentation
+        """
+        Registers the model type in the resource storage manager.
+        :param model_cls: the model to register.
+        :return:
```
ran-z: remove return line

aria/storage/__init__.py
```diff
@@ -353,6 +369,11 @@ def __init__(self, driver, resources=(), **kwargs):
         super(ResourceStorage, self).__init__(driver, resources, **kwargs)
 
     def register(self, resource):
+        """
+        Registers the resource type in the resource storage manager.
+        :param resource: the resource to register.
+        :return:
```
ran-z: remove return line

aria/storage/drivers.py
```diff
@@ -29,7 +29,8 @@
 
 import os
 import shutil
-import distutils.dir_util
+# pylint has an issue with distuls and virtualenvs: https://github.com/PyCQA/pylint/issues/73
```
ran-z: distuls --> distutils?

aria/storage/models.py
```diff
@@ -123,11 +134,12 @@ def __lt__(self, other):
 
         if self.action != other.action:
             if self.action == 'remove':
-                return True
```
ran-z: why was this necessary? what was wrong about the previous implementation?

aria/storage/models.py
```diff
@@ -158,15 +173,18 @@ class DeploymentUpdate(Model):
     deployment_id = Field(type=basestring)
     state = Field(type=basestring, choices=STATES, default=INITIALIZING)
     deployment_plan = Field()
-    deployment_update_nodes = Field(default=None)
-    deployment_update_node_instances = Field(default=None)
+    depup_nodes = Field(default=None)
```
ran-z: why this change?

aria/storage/models.py
```diff
@@ -251,13 +278,21 @@ class Node(Model):
     max_number_of_instances = Field(type=int)
 
     def relationships_by_target(self, target_id):
+        """
+        Retreives all of the relationship by target.
+        :param target_id: the node id of the target  of the relationship
+        :return:
```
ran-z: return statement's empty; since you're yielding it probably requires a different syntax.
i know google's syntax has a ":yields:" keyword, but since we haven't settled on a format yet i dont know if it should be used or simply have the return removed..

aria/storage/models.py
```diff
@@ -297,13 +335,21 @@ class NodeInstance(Model):
     scaling_groups = Field(default=())
 
     def relationships_by_target(self, target_id):
+        """
+        Retreives all of the relationship by target.
+        :param target_id: the instance id of the target  of the relationship
+        :return:
```
ran-z: same as the previous comment

aria/storage/models.py
```diff
@@ -297,13 +335,21 @@ class NodeInstance(Model):
     scaling_groups = Field(default=())
 
     def relationships_by_target(self, target_id):
+        """
+        Retreives all of the relationship by target.
+        :param target_id: the instance id of the target  of the relationship
```
ran-z: double space

aria/storage/structures.py
```diff
@@ -48,6 +48,10 @@
 
 
 def uuid_generator():
+    """
+    wrapper function which generates ids
+    :return:
```
ran-z: empty return

aria/storage/structures.py
```diff
@@ -206,6 +219,10 @@ def __eq__(self, other):
 
     @property
     def fields(self):
+        """
```
ran-z: same as before; plus no capital letter

aria/storage/structures.py
```diff
@@ -174,6 +181,12 @@ class IterPointerField(IterField, PointerField):
 
 
 class Model(object):
+    """
+    Base class for all of the storage models.
+    """
+    # Setting id to avoid missing class member is usage
+    id = None
```
ran-z: should take the comment off. if it makes sense for the "id" to be here then it makes sense, that's it.

where is it used in the base class anyway?

aria/storage/models.py
mxmrlv: too many return statements

aria/storage/models.py
mxmrlv: well, it was to long before we switched to 50, change back?

PR #8, opened 2016-10-20
pylint some top level modules

dankilman:


aria/contexts.py
```diff
@@ -55,6 +61,10 @@ def operation(
             operation_details,
             node_instance,
             inputs=None):
+        """
+        Called during workflow creation, return an operation context. This object is added to the
```
dankilman: This object should be added to the task graph

aria/contexts.py
```diff
@@ -64,61 +74,101 @@ def operation(
...
     def deployment(self):
+        """
+        The deployment model
+        """
         return self.model.deployment.get(self.deployment_id)
 
     @property
     def nodes(self):
+        """
+        Node model iterator
```
dankilman: Iterartor over

PR #10, opened 2016-10-20
pylint-aria-events

mxmrlv:


aria/events/__init__.py
```diff
@@ -13,13 +13,41 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""
+Aria's events Sub-Package
+Path: aria.events
+
+Events package provides events mechanism for different executions in aria.
```
ran-z: ARIA

aria/events/__init__.py
```diff
@@ -13,13 +13,41 @@
...
+"""
+
 import os
 
 from blinker import signal
 
 from ..tools.plugin import plugin_installer
 
 
+__all__ = (
```
ran-z: why was this added?

PR #24, opened 2016-11-16
ARIA-21 Move general code from parser.utils to top level utils

dankilman:


setup.py
```diff
@@ -16,6 +16,7 @@
 
 import os
```
ran-z: think we need to add deployment of CLI binaries/scripts here

PR #40, opened 2017-01-05
ARIA-55 Implement task retry and abort API on model task

dankilman:


aria/orchestrator/workflows/core/events_handler.py
```diff
@@ -43,18 +43,25 @@ def _task_started(task, *args, **kwargs):
...
+        should_retry = all([
+            not isinstance(exception, exceptions.TaskAbortException),
+            task.retry_count < task.max_attempts - 1 or task.max_attempts == task.INFINITE_RETRIES,
+            # ignore_failure check here means the task will not be retries and it will be marked
+            # as failed. The engine will also look at ignore_failure so it won't fail the
+            # workflow.
+            not task.ignore_failure
+        ])
         if should_retry:
+            retry_interval = None
```
ran-z: maybe if this is already initialized to something then it should be the default value or any numeric value rather than None?

dankilman: nope

PR #41, opened 2017-01-05
ARIA-34 Adding get/download resource and render API (as jinja template)

dankilman:


tests/orchestrator/context/test_resource_render.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def test_download_resource_and_render_implicit_ctx_no_variables(tmpdir, ctx):
+    destination = tmpdir.join('destination')
+    ctx.download_resource_and_render(destination=str(destination),
+                                     path=_IMPLICIT_CTX_TEMPLATE_PATH)
+    assert destination.read() == mock.models.DEPLOYMENT_NAME
+
+
+def test_download_resource_and_render_provider_variables(tmpdir, ctx):
```
ran-z: typo (provided) :sweat_smile:

aria/orchestrator/context/common.py
```diff
@@ -97,19 +99,49 @@ def download_resource(self, destination, path=None):
...
+                                             path=path)
+
+    def download_resource_and_render(self, destination, path=None, variables=None):
+        """
+        Download a blueprint resource from the resource storage render its content as a jinja
+        template using the provided variables. ctx is available to the template without providing it
+        explicitly.
+        """
+        self.download_resource(destination=destination, path=path)
+        with open(destination, 'rb') as f:
```
ran-z: why the use of rb and wb? since its rendering doesnt it make sense to use r and w?

dankilman: on windows, omitting the `b` will cause newlines to be written as CRLF regardless of how they were originally. So it is definitely required for `wb`. The `rb` is for consistency

ran-z: got it thanks

PR #42, opened 2017-01-05
ARIA-56 Implement ip property on node instance model

dankilman:


tests/storage/test_models.py
```diff
@@ -664,6 +664,72 @@ def test_node_instance_model_creation(self, node_storage, is_valid, name, runtim
             assert node_instance.deployment == node_storage.deployment.list()[0]
 
 
+class TestNodeInstanceIP(object):
+
+    ip = '1.1.1.1'
+
+    def test_ip_on_none_hosted_node_instance(self, deployment_storage):
```
ran-z: I'd also add a test where the instance in question is not a host but the host does have IP (say as runtime property) and verify the IP is returned.
technically speaking its the same code path as "test_runtime_property_ip_on_hosted_node_instance", but it's a common use case and we should verify it as well

PR #60, opened 2017-01-26
Aria 44 merge parser and storage models

mxmrlv:


extensions/aria_extension_tosca/simple_v1_0/modeling/data_types.py
```diff
@@ -397,8 +397,8 @@ def coerce_value(context, presentation, the_type, entry_schema, constraints, val
 
     if hasattr(the_type, '_coerce_value'):
         # Delegate to '_coerce_value' (likely a DataType instance)
-        return the_type._coerce_value(context, presentation, entry_schema, constraints, value,
-                                      aspect)
+        return the_type._cast_value(context, presentation, entry_schema, constraints, value,
```
tliron: I tried to understand this change and am stumped. I am pretty sure that it's simply a mistake.

aria/storage/modeling/elements.py
```diff
@@ -0,0 +1,102 @@
...
+
+    @property
+    def as_raw(self):
+        return OrderedDict((
+            ('name', self.name),
+            ('type_name', self.type),
+            ('value', self._cast_value()),
+            ('description', self.description)))
+
+    # TODO: change name
```
tliron: What's wrong with the name?

aria/storage/modeling/elements.py
```diff
@@ -0,0 +1,102 @@
...
+    def _cast_value(self):
+        if self.type is None:
+            return
+
+        if self.type.lower() == 'str':
+            return str(self.value)
```
tliron: We absolutely must support Unicode. My suggestion is to use `unicode` by default, even if the type is `str`.

aria/storage/modeling/elements.py
```diff
@@ -0,0 +1,102 @@
...
+        elif self.type.lower() == 'int':
+            return int(self.value)
```
tliron: All these castings might throw exceptions. I think we should catch them and raise better exceptions.

aria/storage/modeling/elements.py
```diff
@@ -0,0 +1,102 @@
...
+        elif self.type.lower() == 'bool':
+            return bool(self.value)
+        elif self.type.lower() == 'float':
+            return float(self.value)
+        else:
+            raise Exception('No supported type_name was provided')
```
tliron: I think we need a better exception type here.

aria/storage/modeling/instance_elements.py
```diff
@@ -0,0 +1,1257 @@
...
+    Properties:
+
+    * :code:`id`: Unique ID (prefixed with the template name)
+    * :code:`type_name`: Must be represented in the :class:`ModelingContext`
+    * :code:`template_name`: Must be represented in the :class:`ServiceModel`
+    * :code:`properties`: Dict of :class:`Parameter`
+    * :code:`interfaces`: Dict of :class:`Interface`
+    * :code:`artifacts`: Dict of :class:`Artifact`
+    * :code:`capabilities`: Dict of :class:`CapabilityTemplate`
+    * :code:`relationship`: List of :class:`Relationship`
```
tliron: Should be plural, `relationships`

aria/storage/modeling/instance_elements.py
```diff
@@ -0,0 +1,1257 @@
...
+            console.puts('Node type: %s' % context.style.type(self.node_type_name))
+            utils.dump_dict_values(context, self.capabilities, 'Capability mappings')
+            utils.dump_dict_values(context, self.requirements, 'Requirement mappings')
+
+
+# endregion
+
+# region Node instances
+
+class NodeBase(structure.ModelMixin):
```
tliron: Does not define `relationships` property

aria/parser/modeling/storage.py
```diff
@@ -0,0 +1,184 @@
...
+    return model.Relationship(
+        source_node=source_node,
+        target_node=target_node)
+
+
+def create_interfaces(context, model_storage, interfaces, obj, attr_name, fn_name):
+    for interface_name, interface in interfaces.iteritems():
+        s_interface = model.Interface(name=interface_name,
+                                      type_name=interface.type_name)
+        setattr(s_interface, attr_name, obj)
```
mxmrlv: why?

aria/parser/modeling/storage.py
```diff
@@ -0,0 +1,184 @@
...
+        service_instance=service_instance)
+
+
+def create_relationship(source_node, target_node):
+    return model.Relationship(
+        source_node=source_node,
+        target_node=target_node)
+
+
+def create_interfaces(context, model_storage, interfaces, obj, attr_name, fn_name):
```
mxmrlv: though this entire module is temporary, obj is kinda rough :)

aria/parser/modeling/storage.py
```diff
@@ -0,0 +1,184 @@
...
+    for interface_name, interface in interfaces.iteritems():
+        s_interface = model.Interface(name=interface_name,
+                                      type_name=interface.type_name)
+        setattr(s_interface, attr_name, obj)
+        try:
+            obj.source_interfaces.append(s_interface)
+        except AttributeError:
+            # is a node
```
mxmrlv: no need to add the interface to the node's interfaces?

aria/orchestrator/workflows/builtin/utils.py
```diff
@@ -68,3 +67,11 @@ def get_task(node_instance_id):
                     graph.add_dependency(dependency, task)
             else:
                 graph.add_dependency(task, dependencies)
+
+
+def _has_operation(interfaces, operation_name):
+    for interface in interfaces:
+        for operation in interface.operations:
```
tliron: I was not aware of that API. By the way, is there a "has_a" API just to check existence? That would be enough.

aria/storage/modeling/instance_elements.py
```diff
@@ -193,7 +193,7 @@ def service_instance(cls):
         return cls.many_to_one_relationship('service_instance')
 
     @declared_attr
-    def interfaces(cls):
+    def interface(cls):
```
tliron: One to many. An interface has zero or more operations, but each operation has only one parent interface.

aria/parser/modeling/storage.py
tliron: I think you misunderstood this function: it's being called both on nodes and on relationships. That's why I called it the generic `obj` and sent in the `attr_name`. If you prefer, we can rename `obj` to `node_or_relationship`. I actually prefer verbose names -- a great example is right now and your misunderstanding of my purpose. :)

aria/parser/modeling/storage.py
tliron: See above -- it works both on nodes and relationships.

aria/parser/modeling/storage.py
tliron: If it's a node it just doesn't have this extra constraint. Line #111 would be enough.

aria/orchestrator/workflows/builtin/utils.py
mxmrlv: yup, thats an sqla syntax. there is a `exists` concept in sqlalcehmy, but from what i've seen it includes importing sqla into the module...
anyway, there's this example:
(ret, ), = Session.query(exists().where(Keyword.name.in_(keywords)))

aria/storage/modeling/instance_elements.py
mxmrlv: but this is a service_instance to interface operation. Since you've changed it from plural to singular, I've assumed that it's a one_to_one relationship.

aria/parser/modeling/storage.py
mxmrlv: actually, i did manage to understand it. (we have a similliar concept in the workflow engine - we chose to call it actor [for the time being]). Anyway, i'm pro verbose as well :).

aria/parser/modeling/storage.py
```diff
@@ -73,7 +75,7 @@ def create_service_template(context):
 def create_service_instance(context, service_template, service_instance_id):
     now = datetime.utcnow()
     return model.ServiceInstance(
-        name='%s_%s' % (service_template.name, service_instance_id),
+        name='{0}_{1}'.format(service_template.name, service_instance_id),
```
mxmrlv: thanks :)

aria/storage/modeling/instance_elements.py
```diff
@@ -264,39 +264,24 @@ def node_fk(cls):
...
-        return cls.many_to_one_relationship('relationship',
-                                            foreign_key_column='relationship_source_fk',
-                                            backreference='source_interfaces')
-
     @declared_attr
-    def relationship_target(cls):
-        return cls.many_to_one_relationship('relationship',
-                                            foreign_key_column='relationship_target_fk',
-                                            backreference='target_interfaces')
+    def relationship(cls):
```
mxmrlv: ok, so there's an issue here. the relationship instance would hold a column "interfaces", this does not imply whether those are source or target interfaces. Only after having a look at each interface we can derive this info according to the edge. 

is this the behavior we want? 
I'm not entirely sure but i could be persuaded :)

aria/storage/modeling/instance_elements.py
mxmrlv: doesn't have to. defined in relationships (as many-to-one)

aria/orchestrator/workflows/api/task.py
```diff
@@ -93,70 +93,133 @@ def __init__(self,
...
-            raise exceptions.TaskException('The operation end should be {0} or {1}'.format(
-                cls.TARGET_OPERATION, cls.SOURCE_OPERATION
-            ))
-        operation_details = getattr(instance.relationship, operation_end)[name]
-        if operation_end == cls.SOURCE_OPERATION:
-            plugins = instance.relationship.source_node.plugins
-            runs_on = model.Task.RUNS_ON_SOURCE
+        assert isinstance(instance, model.Relationship)
+        interface_name = _get_interface_name(name)
+        interfaces = instance.interfaces.filter_by(name=interface_name, edge=edge)
```
mxmrlv: slack: interfaces like or list like?

aria/storage/modeling/elements.py
```diff
@@ -0,0 +1,106 @@
...
+    Represents a typed value.
+
+    This class is used by both service model and service instance elements.
+    """
+    __tablename__ = 'parameter'
+    name = Column(Text, nullable=False)
+    type = Column(Text, nullable=False)
+
+    # Check: value type
+    value = Column(Text)
```
mxmrlv: str_value

aria/storage/modeling/elements.py
```diff
@@ -0,0 +1,106 @@
...
+    @property
+    def as_raw(self):
+        return OrderedDict((
+            ('name', self.name),
+            ('type_name', self.type),
+            ('value', self.raw_value),
+            ('description', self.description)))
+
+    @property
+    def raw_value(self):
```
mxmrlv: value

PR #62, opened 2017-01-30
ARIA-66-Convert-custom-parser-fields-into-sqla-based-fields

mxmrlv:


aria/storage/type.py
```diff
@@ -74,16 +125,115 @@ def coerce(cls, key, value):
...
+
+        return self._strict_map[strict_dict_map_key]
+
+StrictDict = _StrictDict()
+
+
+class _StrictList(object):
+    _strict_map = {}
+
+    def __call__(self, item_cls=None):
```
mxmrlv: doc

aria/storage/type.py
```diff
@@ -175,72 +172,105 @@ class _MutableList(mutable.MutableList):
...
-                return mutable.Mutable.coerce(key, value)
-            else:
-                return value
-
+            return mutable.MutableList.coerce(key, value)
         except ValueError as e:
             raise exceptions.StorageError('SQL Storage error: {0}'.format(str(e)))
 
-StrictDictID = namedtuple('strict_dict_id', 'key_cls, value_cls')
+StrictDictID = namedtuple('StrictDictID', 'key_cls, value_cls')
```
dankilman: `_StrictDictID` and `_StrictValue`

aria/storage/type.py
```diff
@@ -175,72 +172,105 @@ class _MutableList(mutable.MutableList):
...
+            # Creating the type class itself. this class would be returned (used by the sqlalchemy
+            # Column).
             strict_dict_cls = type(
                 'StrictDict_{0}_{1}'.format(key_cls.__name__, value_cls.__name__),
                 (Dict, ),
                 {}
             )
-            type(
+            # Creating the type listening class and associating it with the newly created type.
+            # The new class inherits from both the _MutableList class and the _StrictListMixin,
```
dankilman: wrong comment: not list, dict

aria/storage/type.py
```diff
@@ -175,72 +172,105 @@ class _MutableList(mutable.MutableList):
...
 
-        return self._strict_map[strict_dict_map_key]
+        return self._strict_map[strict_dict_map_key].type_cls
 
 StrictDict = _StrictDict()
 
 
 class _StrictList(object):
+    """
+    This entire class functions as a factory for strict dicts and their listeners.
```
dankilman: wrong comment: not dict, list

aria/storage/type.py
```diff
@@ -175,72 +172,105 @@ class _MutableList(mutable.MutableList):
...
+            # The new class inherits from both the _MutableList class and the _StrictListMixin,
+            # while setting the necessary _key_cls and _value_cls as class attributes.
+            listener_cls = type(
                 'StrictMutableDict_{0}_{1}'.format(key_cls.__name__, value_cls.__name__),
                 (_StrictDictMixin, _MutableDict),
                 {'_key_cls': key_cls, '_value_cls': value_cls}
-            ).associate_with(strict_dict_cls)
-            self._strict_map[strict_dict_map_key] = strict_dict_cls
+            )
+            listener_cls.associate_with(strict_dict_cls)
```
dankilman: shouldn't you not be calling `associate_with` here?

aria/storage/type.py
```diff
@@ -175,72 +172,105 @@ class _MutableList(mutable.MutableList):
...
+        dict((v.type_cls, v.listener_cls) for v in _StrictDict._strict_map.values())
+
+    strict_list_type_to_listener = \
+        dict((v.type_cls, v.listener_cls) for v in _StrictList._strict_map.values())
+
     for prop in mapper.column_attrs:
         column_type = prop.columns[0].type
-        if isinstance(column_type, Dict):
+        # Dict Listeners
+        if any(isinstance(column_type, v) for v in strict_dict_type_to_listener):
```
dankilman: `if type(column_type) in strict_dict_type_to_listener`?
This is the key you are using after all

mxmrlv: well, not really, it's some sqlalcehmy magic, it isn't really an instance of the class (so if was doing something like isintance it would return True, but it doesn't hash the same way [for dicts])

dankilman: I don't understand.
How can `type(column_type) not it strict_dict_type_to_listener` and yet `strict_dict_type_to_listener[type(column_type)].associate_with_attribute`

PR #64, opened 2017-02-01
ARIA-79-concurrent-modifications

dankilman:


aria/storage/base_model.py
```diff
@@ -479,6 +479,7 @@ class NodeInstanceBase(ModelMixin):
     __tablename__ = 'node_instances'
     _private_fields = ['node_fk', 'host_fk']
 
+    version_id = Column(Integer, nullable=False)
```
dankilman: think where version_id might also be applied

aria/orchestrator/workflows/executor/process.py
```diff
@@ -231,6 +186,87 @@ def _update_env(self, env, plugin_prefix):
...
+    def _handle_task_succeeded_request(self, task_id, request, **kwargs):
+        task = self._remove_task(task_id)
+        try:
+            self._apply_tracked_changes(task, request)
+        except BaseException as e:
+            self._task_failed(task, exception=e)
+        else:
+            self._task_succeeded(task)
+
+    def _handle_task_failed_request(self, task_id, request, **kwargs):
```
ran-z: need to decide on semantics - which error is raised?

PR #67, opened 2017-02-03
ARIA-99 Straightforward end-to-end tests for parser and built-in workflow

tliron:


tests/orchestrator/test_complete.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def test_custom():
+    _workflow('maintenance_on')
+
+
+def _workflow(workflow_name):
+    context, _ = consume_node_cellar()
+
+    if workflow_name in BUILTIN_WORKFLOWS:
```
ran-z: this section shouldn't be part of the test
the logic needs to sit elsewhere but the CLI in which case the test can test it as well

tests/orchestrator/test_complete.py
```diff
@@ -0,0 +1,58 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
ran-z: IMO e2e tests need to be separated from unit/functional tests, and sit under, say, "tests/e2e" or something similar.

tests/orchestrator/test_complete.py
```diff
@@ -0,0 +1,58 @@
...
+
+WORKFLOW_POLICY_INTERNAL_PROPERTIES = ('function', 'implementation', 'dependencies')
+
+
+def test_install():
+    _workflow('install')
+
+
+def test_custom():
+    _workflow('maintenance_on')
```
ran-z: I'm not sure we should be testing custom workflows at this time, simply because we haven't quite implemented this feature "for real" yet. It's fine to keep it, I just don't see much benefit for it at the moment.

tests/parser/service_templates.py
```diff
@@ -0,0 +1,30 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from aria.utils.caching import cachedmethod
+
+from .utils import (get_uri, create_context, create_consumer)
+
+
+def consume_node_cellar(consumer_class_name='instance', cache=True):
```
ran-z: sorry for nitpicking but if this module is designated to be a place for several templates to be used by various tests, I think simply "node_cellar" is a better name in this case :)

tests/parser/test_complete.py
```diff
@@ -0,0 +1,40 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
ran-z: I think the module name should be changed, possibly to "test_consumers" or so?

tests/resources/service_templates/node-cellar/types/openstack.yaml
```diff
@@ -0,0 +1,201 @@
...
+
+dsl_definitions:
+
+  openstack:
+    uuid_constraints: &OPENSTACK_UUID_CONSTRAINTS
+      - pattern: '^[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}$'
+
+node_types:
+
+  openstack.Instance:
```
tliron: All these files are just for testing the TOSCA grammar. That's why they are in /tests/. There is no intention for them to be anything other than that.

tests/resources/service_templates/node-cellar/types/aria.yaml
```diff
@@ -0,0 +1,93 @@
+
+policy_types:
```
tliron: To test policy type parsing.

tests/resources/service_templates/node-cellar/node-cellar.yaml
```diff
@@ -0,0 +1,299 @@
...
+
+description: >-
+  Node Cellar TOSCA blueprint.
+  Here is some Unicode: 中國.
+
+metadata:
+  template_name: node-cellar
+  template_author: ARIA
+  template_version: '1.0.0'
+  aria_version: '1.0'
```
tliron: Again, this utterly does not matter. It's here to test metadat parsing, to show that we can parse built in fields as well as custom fields.

tests/resources/service_templates/node-cellar/node-cellar.yaml
```diff
@@ -0,0 +1,299 @@
...
+  default_openstack_credential: &DEFAULT_OPENSTACK_CREDENTIAL
+    user: openstackadmin
+    token: { concat: [ openstack, 123 ] }
+
+repositories:
+
+  node_cellar:
+    description: >-
+      The repository for the Node Cellar application and its dependencies.
+    url: https://github.com/cloudify-cosmo/nodecellar/archive
```
tliron: It could just be a dummy URL.

tests/orchestrator/test_complete.py
tliron: I disagree, but it's not too important. Do you want me to move them? I'd rather not use "e2e" because most people will not know what that means. How about "complete"?

tests/parser/service_templates.py
tliron: I like function names to have verbs in them, because it makes reading code that calls them easier.

tests/orchestrator/test_complete.py
tliron: Since we're rehauling the CLI I am worried about code disappearing. For now I say let's keep this in the test. When we have the new CLI this particular section of code can be redone. I will add a TODO comment here.

tests/orchestrator/test_complete.py
tliron: Since we insist that all our code have tests, I added this.

tests/orchestrator/test_complete.py
ran-z: I don't mind the name much, didn't really mean for it to be "e2e". I think "system_tests" is a better name than "complete", but i'll leave it up to you - I mostly want them separated from the rest - that, I do think is important.

tests/orchestrator/test_complete.py
ran-z: :+1:

tests/complete/test_parser.py
```diff
@@ -0,0 +1,40 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
```
ran-z: Actually, I'm not sure these tests should have moved out from where they were at.
"Complete" means end to end tests; While these are not unit tests per se, they are parser specific.
We already have many tests which aren't standard unit tests yet are treated as such. See execution plugin tests for example - those actually run the workflow engine.
On the other hand, the parser might indeed be used regardless of other components of ARIA by other projects, and so it might be ok to leave this here.
Again, I'll leave it for your consideration.

tests/resources/service_templates/node-cellar/types/aria.yaml
ran-z: lol I understand, but the problem is this yaml import should be part of the extension code and not part of the tests. I realize it's not there right now but I think you might as well have thrown it over there instead right now, no?
Unless I'm missing something, and you think this duplication will remain even once we do have it as part of the extension..?

tests/resources/service_templates/node-cellar/types/openstack.yaml
ran-z: right, if these only test the parser then it makes sense,
what I had in mind was that if we're talking true "complete" tests then in the future we'll want to run tests which use the actual plugin as well etc.

PR #72, opened 2017-03-05
ARIA-105 integrate modeling

tliron:
So, this is a very big pull request.

As it stands, I am not yet done fixing all the tests, however it's been a long time and we need to at least start the review process.

Let me give an overview:

`aria.modeling` is where all the new models sit. They are all very detailed and documented using Sphinx. They ended up diverging quite a lot from what Maxim first created, though I built a lot on his work. In particular, I made sure that SqlAlchemy specifics would be "hidden" from users using standard Python lists and dicts. I reworked a lot of the helper functions in `bases` to support these additional features.

Some highlights:

* A true normalized relational database structure. No fields with string dicts or lists in them.
* No more name lookups: all relationships between models are done using foreign keys. I would actually want to add more constraints in the future to improve validation. As of the now, the `validate` functions are not good enough: a user could create invalid models.
* Parameters are serialized via pickling. This allows us to maintain the rich type structure of TOSCA. So far it seems to be working great, though ... we would need to keep an eye on this.
* There might be seem to be a lot of duplication between templates and instances. This is because 1) the move from one to the other may involve intrinsic function calls, so all properties might indeed be different, and 2) I want to make it possible, in theory at least, to be able to create a service instance even if there is no service template. This would allow the ARIA orchestrator to work on any service, even if it does not have a blueprint. (Actually, a smart modeling tool could possibly "templatize" an existing modeled service, allowing one to generate a blueprint in order to duplicate it. Neat!) The relationship between the two sets of models should be optional, not required.

If you want to see how they are created by the parser, see `aria_extensions_tosca.simple_v1_0.modeling.__init__`.

`aria.orchestrator.workflows.task.api` has been reworked to use both "interface_name" and "operation_name" instead of a single string that is parsed using the period. This has helped a lot in reducing confusion and ambiguity in many parts of the code, and correctly represents the reality of the modeling structure.

[EDIT]: The Task API also intrinsically supports a `dry` argument now used for dry runs. We could think in the future of making this a global default value somewhere in the context -- though it might still be useful to have workflows with some tasks being dry and some wet for debugging purposes.

The impact on our tests has been huge. :/ I am marching through all of them and am not done yet. So please don't be worried to see many tests failing.


aria/cli/args_parser.py
```diff
@@ -137,10 +137,6 @@ def add_workflow_parser(workflow):
         '-w', '--workflow',
         default='install',
         help='The workflow name')
-    workflow.add_argument(
```
ran-z: putting aside the fact this code is going to be thrown out anyway,
how would a workflow execution not require a service instance to run on?

aria/modeling/orchestration.py
```diff
@@ -291,28 +297,35 @@ class ServiceInstanceModificationBase(ModelMixin):
...
-        return cls.foreign_key('service_instance')
+    def service(cls):
+        return cls.many_to_one_relationship('service',
+                                            backreference='modifications')
 
     @declared_attr
-    def service_instance(cls):
-        return cls.many_to_one_relationship('service_instance',
-                                            backreference='modifications')
+    def service_name(cls):
```
ran-z: What's with changing "ServiceInstance" to "Service" all around?

aria/orchestrator/workflows/api/task.py
```diff
@@ -69,21 +70,38 @@ def __init__(self,
...
         :param actor: the operation host on which this operation is registered.
         :param inputs: operation inputs.
         """
-        assert isinstance(actor, (model.Node,
-                                  model.Relationship))
+
+        assert isinstance(actor, (models.Node, models.Relationship))
         super(OperationTask, self).__init__()
+
+        if dry:
```
ran-z: why keep this at all?

aria/orchestrator/workflows/builtin/utils.py
```diff
@@ -14,33 +14,41 @@
 # limitations under the License.
 
 from ..api.task import OperationTask
+from .. import exceptions
 
 
-def create_node_task(operation_name, node):
+def create_node_task(interface_name, operation_name, node, dry=False):
```
ran-z: why is the dry param here now as well?

aria/modeling/types.py
```diff
@@ -23,7 +23,7 @@
 )
 from sqlalchemy.ext import mutable
 
-from .. import exceptions
+from ..storage import exceptions
```
mxmrlv: I think we should create new exceptions - modelling exceptions. from within the same package it think it's better if we trow package specific exceptions.

aria/modeling/utils.py
```diff
@@ -73,7 +46,7 @@ def coerce_value(context, container, value, report_issues=False):
 def validate_dict_values(context, the_dict):
     if not the_dict:
         return
-    validate_list_values(context, the_dict.values())
+    validate_list_values(context, the_dict.itervalues())
```
mxmrlv: python3?

aria/orchestrator/workflows/api/task.py
mxmrlv: why should the task know about whether it's being dry-ran or not?

aria/orchestrator/workflows/api/task.py
```diff
@@ -69,21 +70,38 @@ def __init__(self,
...
+            from ..dry import convert_to_dry
+            plugin, implementation, inputs = convert_to_dry(plugin, implementation, inputs)
+
+        # Coerce inputs
+        if inputs is None:
+            inputs = {}
+        else:
+            for k, v in inputs.iteritems():
```
mxmrlv: python3?

aria/modeling/utils.py
```diff
@@ -120,27 +102,52 @@ def dump_dict_values(context, the_dict, name):
...
+
+
+def copy_locators(target, source):
+    """
+    Copies over :code:`_locator` for all elements, recursively.
+
+    Assumes that target and source have exactly the same list/dict structure.
+    """
+
+    locator = getattr(source, '_locator', None)
```
ran-z: _locator shouldn't be handled in this part of the code

aria/modeling/types.py
```diff
@@ -177,6 +177,7 @@ def coerce(cls, key, value):
         except ValueError as e:
             raise exceptions.StorageError('SQL Storage error: {0}'.format(str(e)))
```
ran-z: i wonder if StorageErrors should be named something different now that the package is named "modeling"

aria/modeling/misc.py
```diff
@@ -0,0 +1,232 @@
...
+
+    @value.setter
+    def value(self, value):
+        if value is None:
+            self._value = None
+        else:
+            try:
+                self._value = pickle.dumps(value)
+            except (pickle.PicklingError, TypeError):
+                logging.getLogger('aria').warn('Could not pickle parameter of type "{0}": {1}'
```
ran-z: this needs to raise an error, not log a warning

aria/cli/args_parser.py
tliron: It's in the context. Awkward, I know. :/ This is something we should fix outside the scope of this JIRA.

aria/orchestrator/workflows/api/task.py
```diff
@@ -69,21 +73,38 @@ def __init__(self,
...
+                if not isinstance(v, models.Parameter):
+                    inputs[k] = models.Parameter(name=k,
+                                                 type_name=full_type_name(v),
+                                                 value=v)
+
+        self.name = name
         self.actor = actor
-        self.name = '{name}.{actor.id}'.format(name=name, actor=actor)
```
ran-z: why was this changed?

aria/orchestrator/workflows/api/task.py
```diff
@@ -69,21 +73,38 @@ def __init__(self,
...
+        if dry:
+            from ..dry import convert_to_dry
+            plugin, implementation, inputs = convert_to_dry(plugin, implementation, inputs)
+
+        # Coerce inputs
+        if inputs is None:
+            inputs = {}
+        else:
+            for k, v in inputs.iteritems():
+                if not isinstance(v, models.Parameter):
```
ran-z: im not sure i understand why this check is needed

aria/orchestrator/workflows/core/task.py
```diff
@@ -66,65 +67,60 @@ class StubTask(BaseTask):
...
-        elif isinstance(api_task.actor, model.Relationship):
+            task_model_cls = base_task_model.as_node_task
+        elif isinstance(api_task.actor, models.Relationship):
             context_cls = operation_context.RelationshipOperationContext
-            task_model_cls = base_task_model.as_relationship_instance
+            task_model_cls = base_task_model.as_relationship_task
         else:
             raise RuntimeError('No operation context could be created for {actor.model_cls}'
                                .format(actor=api_task.actor))
-        plugin = api_task.plugin
```
ran-z: why was this deleted again?

aria/modeling/service_template.py
```diff
@@ -0,0 +1,1693 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
ran-z: shouldnt this module be called templates and not service templates?

aria/orchestrator/workflows/api/task.py
```diff
@@ -69,21 +73,38 @@ def __init__(self,
...
+                    inputs[k] = models.Parameter(name=k,
+                                                 type_name=full_type_name(v),
```
ran-z: the type here should in principle correspond to the one in template's operation's inputs, correct? i understand theres no validation on this at the moment but im asking whether theres any benefit to saving the type name of an arbitrary input

aria/cli/commands.py
```diff
@@ -230,43 +228,48 @@ def _parse(self, uri):
     
     def _get_workflow(self, context, workflow_name):
         if workflow_name in BUILTIN_WORKFLOWS:
-            workflow_fn = import_fullname('aria.orchestrator.workflows.builtin.%s' % workflow_name)
+            workflow_fn = import_fullname('aria.orchestrator.workflows.builtin.{0}'.format(
+                workflow_name))
             inputs = {}
         else:
+            workflow = None
+            for policy in context.modeling.instance.policies:
```
ran-z: if by now there should be a service in the storage, shouldn't this access be to the service's "operations" field instead?

aria/modeling/service.py
```diff
@@ -0,0 +1,1529 @@
...
+        return satisfied
+
+    def validate_capabilities(self, context):
+        satisfied = True
+        for node in self.nodes:
+            if not node.validate_capabilities(context):
+                satisfied = False
+        return satisfied
+
+    def find_nodes(self, node_template_name):
```
mxmrlv: couldn't this be replaced by an sql query (in a single line), and if so, why do we need this helper method?

aria/modeling/service.py
```diff
@@ -0,0 +1,1529 @@
...
+        nodes = []
+        for node in self.nodes:
+            if node.node_template.name == node_template_name:
+                nodes.append(node)
+        return collections.FrozenList(nodes)
+
+    def get_node_ids(self, node_template_name):
```
mxmrlv: ditto

aria/modeling/service.py
```diff
@@ -0,0 +1,1529 @@
...
+        for group in self.groups:
+            if group.template_name == group_template_name:
+                groups.append(group)
+        return collections.FrozenList(groups)
+
+    def get_group_ids(self, group_template_name):
+        return collections.FrozenList((group.name
+                                       for group in self.find_groups(group_template_name)))
+
+    def is_node_a_target(self, context, target_node):
```
mxmrlv: everynode which is a target has inbound_relationships. these methods are probably not relavent anymore

aria/modeling/service.py
```diff
@@ -0,0 +1,1529 @@
...
+        utils.coerce_list_values(context, container, self.nodes, report_issues)
+        utils.coerce_list_values(context, container, self.groups, report_issues)
+        utils.coerce_list_values(context, container, self.policies, report_issues)
+        if self.substitution is not None:
+            self.substitution.coerce_values(context, container, report_issues)
+        utils.coerce_dict_values(context, container, self.inputs, report_issues)
+        utils.coerce_dict_values(context, container, self.outputs, report_issues)
+        utils.coerce_dict_values(context, container, self.operations, report_issues)
+
+    def dump(self, context):
```
mxmrlv: what does this method does?

aria/modeling/service.py
```diff
@@ -0,0 +1,1529 @@
...
+            # Find target template
+            target_node_template, target_node_capability = \
+                requirement_template.find_target(context, node_template)
+            if target_node_template is not None:
+                satisfied = self._satisfy_capability(context,
+                                                     target_node_capability,
+                                                     target_node_template,
+                                                     requirement_template)
+            else:
+                context.validation.report('requirement "{0}" of node "{1}" has no target node '
```
mxmrlv: there should be unified model for logging, it's weird that some log msgs, while other report it to the parser's context...

aria/modeling/service_template.py
```diff
@@ -0,0 +1,1693 @@
...
+            ('node_templates', formatting.as_raw_list(self.node_templates)),
+            ('group_templates', formatting.as_raw_list(self.group_templates)),
+            ('policy_templates', formatting.as_raw_list(self.policy_templates)),
+            ('substitution_template', formatting.as_raw(self.substitution_template)),
+            ('inputs', formatting.as_raw_dict(self.inputs)),
+            ('outputs', formatting.as_raw_dict(self.outputs)),
+            ('operation_templates', formatting.as_raw_list(self.operation_templates))))
+
+    @property
+    def types_as_raw(self):
```
mxmrlv: is this really needed?

aria/modeling/service_template.py
```diff
@@ -0,0 +1,1693 @@
...
+            group_template.dump(context)
+        for policy_template in self.policy_templates:
+            policy_template.dump(context)
+        if self.substitution_template is not None:
+            self.substitution_template.dump(context)
+        utils.dump_dict_values(context, self.inputs, 'Inputs')
+        utils.dump_dict_values(context, self.outputs, 'Outputs')
+        utils.dump_dict_values(context, self.operation_templates, 'Operation templates')
+
+    def dump_types(self, context):
```
mxmrlv: this as well

aria/modeling/service_template.py
mxmrlv: How about templates.py and instances.py?

aria/modeling/service.py
```diff
@@ -0,0 +1,1529 @@
...
+        return cls.many_to_many_relationship('parameter', table_prefix='inputs',
+                                             dict_key='name')
+
+    @declared_attr
+    def outputs(cls):
+        return cls.many_to_many_relationship('parameter', table_prefix='outputs',
+                                             dict_key='name')
+
+    @declared_attr
+    def operations(cls):
```
ran-z: change this to "workflows"

aria/modeling/service.py
```diff
@@ -0,0 +1,1529 @@
...
+        return collections.FrozenList((group.name
+                                       for group in self.find_groups(group_template_name)))
+
+    def is_node_a_target(self, context, target_node):
+        for node in self.nodes:
+            if self._is_node_a_target(context, node, target_node):
+                return True
+        return False
+
+    def _is_node_a_target(self, context, source_node, target_node):
```
ran-z: need to get rid of context throughout the models..

aria/modeling/service.py
```diff
@@ -0,0 +1,1529 @@
...
+                          'service_template_fk']
+
+    # Service one-to-one to Substitution
+    @declared_attr
+    def substitution_fk(cls):
+        return cls.foreign_key('substitution', nullable=True)
+
+    # Service many-to-one to ServiceTemplate
+    @declared_attr
+    def service_template_fk(cls):
```
ran-z: service_template.services should return a dictionary (names as the keys) and not a list - same goes for all models with a unique name.

aria/modeling/orchestration.py
```diff
@@ -327,20 +340,48 @@ class PluginBase(ModelMixin):
     uploaded_at = Column(DateTime, nullable=False, index=True)
     wheels = Column(List, nullable=False)
 
+    # region foreign keys
+
+    __private_fields__ = ['service_template_fk']
```
ran-z: as we mentioned, this should change; need two separate plugin models, one for those declared in service templates, and another for those installed via the plugin manager.

aria/modeling/service_changes.py
```diff
@@ -0,0 +1,219 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
ran-z: why create this..?

aria/orchestrator/workflows/api/task.py
```diff
@@ -165,13 +166,21 @@ def for_relationship(cls, relationship, interface_name, operation_name, inputs=N
                 'Could not find operation "{0}" on interface "{1}" for relationship "{2}"'.format(
                     operation_name, interface_name, relationship.name))
 
+        plugin = None
```
ran-z: refactor?

aria/modeling/bases.py
```diff
@@ -360,7 +361,7 @@ def __repr__(self):
 
 class ModelIDMixin(object):
     id = Column(Integer, primary_key=True, autoincrement=True)
-    name = Column(Text, nullable=True, index=True)
+    name = Column(Text, index=True)
```
mxmrlv: why?

aria/modeling/misc.py
```diff
@@ -42,11 +43,11 @@ class ParameterBase(TemplateModelMixin):
 
     __tablename__ = 'parameter'
 
-    name = Column(Text, nullable=False)
-    type_name = Column(Text, nullable=False)
+    name = Column(Text)
```
mxmrlv: it doesn't make sense to require for each model to have a name...

aria/modeling/service_template.py
```diff
@@ -97,6 +97,34 @@ def outputs(cls):
     def operation_templates(cls):
         return cls.one_to_many_relationship('operation_template', dict_key='name')
 
+    @declared_attr
+    def node_types(cls):
+        return cls.one_to_one_relationship('type', key='node_type_fk', backreference='')
```
mxmrlv: this solution looks weird. Having to pass blank string in order not to have backref...

aria/orchestrator/workflows/api/task.py
```diff
@@ -173,6 +170,13 @@ def for_relationship(cls, relationship, interface_name, operation_name, inputs=N
             **kwargs)
 
     @classmethod
+    def _get_operation(cls, interfaces, interface_name, operation_name):
```
mxmrlv: staticmethod?

aria/modeling/service_instance.py
```diff
@@ -192,21 +193,21 @@ def get_group_ids(self, group_template_name):
...
+            if self._is_node_a_target(node, target_node):
                 return True
         return False
 
-    def _is_node_a_target(self, context, source_node, target_node):
-        if source_node.relationships:
-            for relationship in source_node.relationships:
-                if relationship.target_node_id == target_node.name:
+    def _is_node_a_target(self, source_node, target_node):
+        if source_node.outbound_relationships:
```
mxmrlv: if source_node.outbound_relationships.filter_by(target_node_name=target_node.name):...

tests/modeling/test_models.py
```diff
@@ -173,92 +168,83 @@ def _test_model(is_valid, storage, model_cls, model_kwargs):
...
-        'is_valid, plan, description, created_at, updated_at, main_file_name',
+        'is_valid, description, created_at, updated_at, main_file_name',
         [
-            (False, None, 'description', now, now, '/path'),
-            (False, {}, {}, now, now, '/path'),
-            (False, {}, 'description', 'error', now, '/path'),
-            (False, {}, 'description', now, 'error', '/path'),
-            (False, {}, 'description', now, now, {}),
-            (True, {}, 'description', now, now, '/path'),
+            (True, 'description', now, now, '/path'),
```
mxmrlv: group False and True together?

tests/modeling/test_models.py
```diff
@@ -486,83 +472,79 @@ def test_deployment_update_step_order(self):
...
-            (True, 'name', 1, 1, 1, None, [], 'type', []),
-            (True, 'name', 1, 1, 1, [], [], 'type', None),
+            (False, m_cls, 1, 1, 1, {}, {}),
+            (False, 'name', m_cls, 1, 1, {}, {}),
+            (False, 'name', 1, m_cls, 1, {}, {}),
+            (False, 'name', 1, 1, m_cls, {}, {}),
+            (False, 'name', 1, 1, 1, m_cls, {}),
+            (False, 'name', 1, 1, 1, None, {}),
+
+            (True, 'name', 1, 1, 1, {}, {}),
```
mxmrlv: what happened to the rest of the valid node template creation params?

tests/modeling/test_models.py
```diff
@@ -663,66 +646,45 @@ def _node(self, storage, node, is_host, ip, host_fk=None):
...
-             0),
-            (True, 0, None, 'target_cap_name', 'type_name', 'template_name', [], 0, 0),
-            (True, 0, 'target_node_id', None, 'type_name', 'template_name', [], 0, 0),
-            (True, 0, 'target_node_id', 'target_cap_name', None, 'template_name', [], 0, 0),
-            (True, 0, 'target_node_id', 'target_cap_name', 'type_name', None, [], 0, 0),
-            (True, 0, 'target_node_id', 'target_cap_name', 'type_name', 'template_name', [], None,
-             0),
-            (True, 0, 'target_node_id', 'target_cap_name', 'type_name', 'template_name', [], 0,
-             None),
+            (False, m_cls, 0),
```
mxmrlv: what happened to the rest of the test cases?

tests/orchestrator/workflows/executor/test_executor.py
```diff
@@ -43,7 +43,9 @@ def test_execute(executor):
     expected_value = 'value'
     successful_task = MockTask(mock_successful_task)
     failing_task = MockTask(mock_failing_task)
-    task_with_inputs = MockTask(mock_task_with_input, inputs=dict(input='value'))
+    #task_with_inputs = MockTask(mock_task_with_input, inputs=dict(input='value'))
```
mxmrlv: leftovers

aria/orchestrator/workflows/api/task.py
```diff
@@ -92,144 +93,133 @@ def __init__(self,
...
-        final_inputs = dict((p.name, p.as_raw['value']) for p in operation_inputs)
-        final_inputs.update(additional_inputs or {})
-        return final_inputs
+        # Wrap inputs
+        if inputs:
+            for k, v in inputs.iteritems():
+                if not isinstance(v, models.Parameter):
+                    inputs[k] = models.Parameter.wrap(k, v)
+
+        # TODO: These extra inputs should likely be stored as a separate entry in the task model,
```
ran-z: i'm not sure I agree with this comment, but we can discuss it at a later time

aria/orchestrator/workflows/api/task.py
```diff
@@ -92,144 +93,133 @@ def __init__(self,
...
+                        .format(operation_name, interface_name, actor_type, actor.name))
+
+            self.implementation = operation.implementation
+            self.inputs = OperationTask._merge_inputs(operation.inputs, inputs)
+
+            self.name = OperationTask.NAME_FORMAT.format(type=actor_type,
+                                                         name=actor.name,
+                                                         interface=interface_name,
+                                                         operation=operation_name)
+        else:
```
ran-z: what is this for?

tests/orchestrator/workflows/executor/test_executor.py
tliron: Fixed

tests/modeling/test_models.py
tliron: `Relationship` comprises almost entirely foreign keys now! For the test I left the only few columns that are values.

tests/modeling/test_models.py
tliron: Fixed

aria/orchestrator/workflows/builtin/utils.py
```diff
@@ -14,33 +14,38 @@
...
 
-    if _has_operation(node.interfaces, operation_name):
-        return OperationTask.node(instance=node,
-                                  name=operation_name)
-    return None
+    try:
+        return OperationTask.for_node(node=node,
+                                      interface_name=interface_name,
+                                      operation_name=operation_name)
+    except exceptions.TaskException:
```
tliron: OK. This means that I will need to create a new exception sub-class here, specifically for task creation.

aria/orchestrator/workflows/api/task.py
```diff
@@ -92,144 +93,133 @@ def __init__(self,
...
+
+    @staticmethod
+    def _get_operation(interfaces, interface_name, operation_name):
+        interface = interfaces.get(interface_name)
+        if interface is not None:
+            return interface.operations.get(operation_name)
+        return None
+
+    @staticmethod
+    def _find_plugin(plugin_specification):
```
tliron: I'm prepared to hear a better suggestion as to where to get the list of installed plugins from.

aria/orchestrator/workflows/api/task.py
tliron: Changed the comment to say "suggestion"

aria/orchestrator/workflows/api/task.py
tliron: We have quite a few tests that instantiate `OperationTask` "from scratch" using raw arguments, instead of calling `for_node` and `for_relationship`. In order to support those I needed to allow for this separate codepath.

tests/modeling/test_models.py
tliron: Since moving so much to foreign keys, there are much fewer raw columns to test now.

aria/modeling/service_instance.py
```diff
@@ -0,0 +1,1553 @@
...
+    @declared_attr
+    def service_template(cls):
+        return relationships.many_to_one(cls, 'service_template')
+
+    description = Column(Text)
+
+    @declared_attr
+    def meta_data(cls):
+        # Warning! We cannot use the attr name "metadata" because it's used by SQLAlchemy!
+        return relationships.many_to_many(cls, 'metadata', dict_key='name')
```
tliron: `meta_data`. Not sure I understand the question?

aria/modeling/service_instance.py
```diff
@@ -0,0 +1,1553 @@
...
+                    self.target_capability.name)))
+            if self.type is not None:
+                console.puts('Relationship type: {0}'.format(context.style.type(self.type.name)))
+            if (self.relationship_template is not None) and self.relationship_template.name:
+                console.puts('Relationship template: {0}'.format(
+                    context.style.node(self.relationship_template.name)))
+            utils.dump_dict_values(self.properties, 'Properties')
+            utils.dump_interfaces(self.interfaces, 'Interfaces')
+
+    __private_fields__ = ['type_fk',
```
tliron: I always prefer everything private to be pushed to the end, because it's an implementation detail. People reading the code want to see what's public first. But I will change it to your preference.

aria/modeling/relationships.py
```diff
@@ -0,0 +1,402 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
tliron: Well, we usually name our modules in plural. But OK, I will change this.

aria/modeling/orchestration.py
```diff
@@ -0,0 +1,350 @@
...
+        if value < 1 and value != TaskBase.INFINITE_RETRIES:
+            raise ValueError('Max attempts can be either -1 (infinite) or any positive number. '
+                             'Got {value}'.format(value=value))
+        return value
+
+    # region foreign keys
+
+    @declared_attr
+    def node_fk(cls):
+        return relationships.fk('node', nullable=True)
```
tliron: Will change

aria/modeling/models.py
```diff
@@ -0,0 +1,283 @@
...
+    PluginSpecification,
+
+    # Orchestration models
+    Execution,
+    Plugin,
+    Task,
+    Log
+]
+
+__all__ = (
```
tliron: I wanted this to be close to `models_to_register` because they are almost the same, but I will do it your way. I am adding comments though so people will know to update the other list.

aria/cli/dry.py
```diff
@@ -0,0 +1,88 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
tliron: We said CLI.

aria/modeling/__init__.py
```diff
@@ -0,0 +1,48 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
tliron: These methods have almost no use on their own. They are utility methods that are exactly only useful for the complete instantiation of the service template. I suggest we wait until we have an instantiation (or however we will call it) module.

aria/orchestrator/workflows/api/task.py
```diff
@@ -92,144 +93,133 @@ def __init__(self,
...
+            operation_name=operation_name,
+            max_attempts=max_attempts,
+            retry_interval=retry_interval,
+            ignore_failure=ignore_failure,
+            inputs=inputs,
+            runs_on=runs_on)
+
+    @staticmethod
+    def _get_operation(interfaces, interface_name, operation_name):
+        interface = interfaces.get(interface_name)
```
tliron: Done

aria/modeling/relationships.py
```diff
@@ -0,0 +1,402 @@
...
+            other_property = model_class.__tablename__
+        backref_kwargs = backref_kwargs or {}
+        return relationship(
+            lambda: _get_class_for_table(model_class, other_table),
+            backref=backref(other_property, **backref_kwargs),
+            **relationship_kwargs
+        )
+
+
+def _get_class_for_table(model_class, tablename):
```
tliron: Right, it should be opposite. Fixed.

tliron: It receives both a class and a tablename, and returns a class. This was your function originally, I did not change it.

aria/orchestrator/workflows/builtin/utils.py
```diff
@@ -51,14 +56,14 @@ def create_node_task_dependencies(graph, tasks_and_nodes, reverse=False):
 
     def get_task(node_id):
```
tliron: Since you will be fixing the builtins in ARIA-120, I want to suggest that perhaps much of these needs to be reworked, and it is better to be reworked there.

PR #74, opened 2017-03-05
ARIA-115-Log-model-should-have-an-Execution-field

mxmrlv:


aria/storage/modeling/orchestrator_elements.py
```diff
@@ -471,6 +471,14 @@ def retry(message=None, retry_interval=None):
 class LogBase(ModelMixin):
     __tablename__ = 'log'
 
+    @declared_attr
+    def execution_fk(cls):
+        return cls.foreign_key('execution', nullable=True)
```
ran-z: nullable should be False

PR #75, opened 2017-03-07
ARIA-117-Log-model-should-have-an-Task-field

mxmrlv:


aria/orchestrator/context/operation.py
```diff
@@ -64,11 +67,19 @@ def task(self):
...
+        # SQLAlchemy prevents from accessing an object which was created on a different thread.
+        # So we retrieve the object from the storage if the current thread isn't the same as the
+        # original thread.
+        if not self._task or self._current_thread_id != self._current_thread:
             self._task = self.model.task.get(self._task_id)
+            self._current_thread = self._current_thread_id
         return self._task
 
     @property
+    def _current_thread_id(self):
```
ran-z: be a man. save it in a thread local.

aria/orchestrator/execution_plugin/common.py
```diff
@@ -133,14 +133,14 @@ def check_error(ctx, error_check_func=None, reraise=False):
     _error = ctx._error
     # this happens when a script calls task.abort/task.retry more than once
     if isinstance(_error, RuntimeError):
-        ctx.task._original_abort(str(_error))
```
ran-z: check if the original abort/retry are still needed..?

PR #81, opened 2017-03-22
ARIA-83 Support ARIA profile for TOSCA

tliron:
Notes:

* The TOSCA parser extension now supports "special" imports, in this case the name is `aria-1.0`. This allows service template writers to just use the name without having to know where it's located.
* Clarified, documented, and cleaned up models and other code having to do with plugin specifications and workflows.
* Fixed all relevant tests.


aria/cli/commands.py
```diff
@@ -241,12 +241,9 @@ def _get_workflow(self, context, workflow_name):
             if workflow.type.role != 'workflow':
                 raise AttributeError('policy is not a workflow: "{0}"'.format(workflow_name))
 
-            try:
-                sys.path.append(workflow.properties['implementation'].value)
-            except KeyError:
-                pass
+            sys.path.append(os.path.dirname(str(context.presentation.location)))
```
ran-z: why was this changed?

tliron: My previous implementation had a stupid extra `function` property which was just a temporary solution. Obviously it's enough to just have an `implementation` property. Until we handle true CSAR, I guess we can just pretend that artifacts are in the same directory as the main YAML.

ran-z: I actually didn't mean the `function` part, but rather the part where we append the directory of the main yaml rather than a separate implementation string.
workflows code should be treated the same as operations code i.e. it comes from plugins, at the moment we dont install it via csar so it made more sense to me to keep it the first way but i guess for the time being it'll indeed be more convenient to supply workflows this way - however it means that if you install workflow code with your plugins you wont be able to run it.

PR #91, opened 2017-03-31
ARIA-134 Parser populates service workflows

tliron:
Also involved some small fixes to service models.

### DIFF
--- a/extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
+++ b/extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
@@ -85,9 +85,14 @@ def create_service_template_model(context): # pylint: disable=too-many-locals,to
     policies = context.presentation.get('service_template', 'topology_template', 'policies')
     if policies:
         for policy in policies.itervalues():
-            if model.policy_types.get_descendant(policy.type).role == 'plugin':
+            role = model.policy_types.get_descendant(policy.type).role
+            if role == 'plugin':
                 plugin_specification = create_plugin_specification_model(context, policy)
                 model.plugin_specifications[plugin_specification.name] = plugin_specification
+            elif role == 'workflow':
+                operation_template = create_workflow_operation_template_model(context,
+                                                                              model, policy)
+                model.workflow_templates[operation_template.name] = operation_template
 
     # Node templates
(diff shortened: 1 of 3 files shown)


extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
```diff
@@ -446,6 +451,29 @@ def get(name):
     return model
 
 
+def create_workflow_operation_template_model(context, service_template, policy):
+    model = OperationTemplate(name=policy._name,
```
ran-z: Do we want workflows to be modeled as OperationTemplates?
There are indeed many similarities though some fields don't really apply, e.g. `retry_interval`; What do you think?

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
```diff
@@ -85,9 +85,14 @@ def create_service_template_model(context): # pylint: disable=too-many-locals,to
     policies = context.presentation.get('service_template', 'topology_template', 'policies')
     if policies:
         for policy in policies.itervalues():
-            if model.policy_types.get_descendant(policy.type).role == 'plugin':
+            role = model.policy_types.get_descendant(policy.type).role
```
ran-z: what about policies which are neither plugin or workflows?

I think the expected behavior should be that any such policy will be available under `service.policies`, while the plugin and workflow ones won't be available under `service.policies`, no?

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
tliron: You are not wrong ... but I also wanted to minimize the number of models we have. I think if we do decide to change this, we can do it in a separate JIRA.

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
tliron: *All* the policies are added in line 115, below. Well, I think we *have* to add all of them because they are there in the TOSCA service template. The fact that ARIA recognizes some special roles doesn't mean they aren't policies, still.

PR #93, opened 2017-04-02
ARIA-136-ctx-binary-doesnt-get-installed-via-pip-install

mxmrlv:


setup.py
```diff
@@ -63,20 +64,43 @@
...
+    custom_cmd.skip_ctx = False
+
+
+def _run(custom_cmd):
+    if custom_cmd.skip_ctx is False:
+        console_scripts.append('ctx = aria.orchestrator.execution_plugin.ctx_proxy.client:main')
+    custom_cmd.command.run(custom_cmd)
+
+
+class InstallCommand(install):
```
ran-z: why not have a base CustomCommand class,
have all the private functions in it as well,
and have two subclasses which only set `command`?

setup.py
```diff
@@ -63,20 +64,43 @@
...
 
-    def run(self):
-        if self.skip_ctx is False:
-            console_scripts.append('ctx = aria.orchestrator.execution_plugin.ctx_proxy.client:main')
-        install.run(self)
+def _generate_boolean_options(command):
+    return command.boolean_options + ['skip-ctx']
+
+
+def _initialize_options(custom_cmd):
```
ran-z: who's actually calling this?

setup.py
mxmrlv: Thats where is was coming from at first. But the problem is with the class attributes...

PR #95, opened 2017-04-04
ARIA-92 Automatic operation task configuration

tliron:
Main changes:

1. Removed `runs_on` field from task model and API
2. Two new instantiations phase, one to find the host nodes, and the second to configure operation instances for the execution plugin
3. Parser fills in `configuration` field in `OperationTemplate` using `dependencies` in TOSCA syntax
4. Fixes to reqs-and-caps bugs

### DIFF
--- a/aria/modeling/orchestration.py
+++ b/aria/modeling/orchestration.py
@@ -67,13 +67,13 @@ class ExecutionBase(ModelMixin):
     CANCELLING = 'cancelling'
     FORCE_CANCELLING = 'force_cancelling'
 
-    STATES = [TERMINATED, FAILED, CANCELLED, PENDING, STARTED, CANCELLING, FORCE_CANCELLING]
-    END_STATES = [TERMINATED, FAILED, CANCELLED]
+    STATES = (TERMINATED, FAILED, CANCELLED, PENDING, STARTED, CANCELLING, FORCE_CANCELLING)
+    END_STATES = (TERMINATED, FAILED, CANCELLED)
 
     VALID_TRANSITIONS = {
-        PENDING: [STARTED, CANCELLED],
-        STARTED: END_STATES + [CANCELLING],
-        CANCELLING: END_STATES + [FORCE_CANCELLING]
+        PENDING: (STARTED, CANCELLED),
+        STARTED: END_STATES + (CANCELLING,),
+        CANCELLING: END_STATES + (FORCE_CANCELLING,)
     }
 
     @orm.validates('status')
@@ -219,7 +219,44 @@ class PluginBase(ModelMixin):
 
 class TaskBase(ModelMixin):
     """
-    A Model which represents an task
+    Represents the smallest unit of stateful execution in ARIA. The task state includes inputs,
+    outputs, as well as an atomic status, ensuring that the task can only be running once at any
+    given time.
+
+    Tasks may be "one shot" or may be configured to run repeatedly in the case of failure.
+
+    Tasks are often based on :class:`Operation`, and thus act on either a :class:`Node` or a
+    :class:`Relationship`, however this is not required.
+
+    :ivar node: The node actor (optional)
+    :vartype node: :class:`Node`
+    :ivar relationship: The relationship actor (optional)
+    :vartype relationship: :class:`Relationship`
+    :ivar plugin: The implementing plugin (set to None for default execution plugin)
+    :vartype plugin: :class:`Plugin`
+    :ivar inputs: Parameters that can be used by this task
+    :vartype inputs: {basestring: :class:`Parameter`}
+    :ivar implementation: Python path to an ``@operation`` function
+    :vartype implementation: basestring
+    :ivar max_attempts: Maximum number of retries allowed in case of failure
+    :vartype max_attempts: int
+    :ivar retry_interval: Interval between retries (in seconds)
+    :vartype retry_interval: int
+    :ivar ignore_failure: Set to True to ignore failures
+    :vartype ignore_failure: bool
+    :ivar due_at: Timestamp to start the task
+    :vartype due_at: datetime
+    :ivar execution: Assigned execution
+    :vartype execution: :class:`Execution`
+    :ivar status: Current atomic status ('pending', 'retrying', 'sent', 'started', 'success',
+                  'failed')
+    :vartype status: basestring
+    :ivar started_at: Timestamp for when task started
+    :vartype started_at: datetime
+    :ivar ended_at: Timestamp for when task ended
+    :vartype ended_at: datetime
+    :ivar retry_count: How many retries occurred
+    :vartype retry_count: int
     """
 
     __tablename__ = 'task'
@@ -227,7 +264,7 @@ class TaskBase(ModelMixin):
     __private_fields__ = ['node_fk',
                           'relationship_fk',
                           'plugin_fk',
-                          'execution_fk',
+                          'execution_fk'
                           'node_name',
                           'relationship_name',
                           'execution_name']
@@ -246,11 +283,6 @@ class TaskBase(ModelMixin):
         SUCCESS,
         FAILED,
     )
-
-    RUNS_ON_SOURCE = 'source'
-    RUNS_ON_TARGET = 'target'
-    RUNS_ON_NODE = 'node'
-    RUNS_ON = (RUNS_ON_NODE, RUNS_ON_SOURCE, RUNS_ON_TARGET)
 
     INFINITE_RETRIES = -1
 
@@ -278,37 +310,25 @@ class TaskBase(ModelMixin):
     def inputs(cls):
         return relationship.many_to_many(cls, 'parameter', prefix='inputs', dict_key='name')
 
-    status = Column(Enum(*STATES, name='status'), default=PENDING)
-
-    due_at = Column(DateTime, nullable=False, index=True, default=datetime.utcnow())
-    started_at = Column(DateTime, default=None)
-    ended_at = Column(DateTime, default=None)
+    implementation = Column(String)
     max_attempts = Column(Integer, default=1)
-    retry_count = Column(Integer, default=0)
     retry_interval = Column(Float, default=0)
     ignore_failure = Column(Boolean, default=False)
 
-    # Operation specific fields
-    implementation = Column(String)
-    _runs_on = Column(Enum(*RUNS_ON, name='runs_on'), name='runs_on')
+    # State
+    status = Column(Enum(*STATES, name='status'), default=PENDING)
+    due_at = Column(DateTime, nullable=False, index=True, default=datetime.utcnow())
+    started_at = Column(DateTime, default=None)
+    ended_at = Column(DateTime, default=None)
+    retry_count = Column(Integer, default=0)
 
     @property
     def has_ended(self):
-        return self.status in [self.SUCCESS, self.FAILED]
+        return self.status in (self.SUCCESS, self.FAILED)
 
     @property
     def is_waiting(self):
-        return self.status in [self.PENDING, self.RETRYING]
-
-    @property
-    def runs_on(self):
-        if self._runs_on == self.RUNS_ON_NODE:
-            return self.node
-        elif self._runs_on == self.RUNS_ON_SOURCE:
-            return self.relationship.source_node  # pylint: disable=no-member
-        elif self._runs_on == self.RUNS_ON_TARGET:
-            return self.relationship.target_node  # pylint: disable=no-member
-        return None
+        return self.status in (self.PENDING, self.RETRYING)
 
     @property
     def actor(self):
@@ -366,12 +386,12 @@ class TaskBase(ModelMixin):
     # endregion
 
     @classmethod
-    def for_node(cls, instance, runs_on, **kwargs):
-        return cls(node=instance, _runs_on=runs_on, **kwargs)
+    def for_node(cls, actor, **kwargs):
+        return cls(node=actor, **kwargs)
 
     @classmethod
-    def for_relationship(cls, instance, runs_on, **kwargs):
-        return cls(relationship=instance, _runs_on=runs_on, **kwargs)
+    def for_relationship(cls, actor, **kwargs):
+        return cls(relationship=actor, **kwargs)
 
     @staticmethod
     def abort(message=None):
--- a/aria/modeling/service_instance.py
+++ b/aria/modeling/service_instance.py
@@ -20,12 +20,14 @@ from sqlalchemy import (
     Text,
     Integer,
     Enum,
+    Boolean
 )
 from sqlalchemy import DateTime
 from sqlalchemy.ext.associationproxy import association_proxy
 from sqlalchemy.ext.declarative import declared_attr
 
 from .mixins import InstanceModelMixin
+from ..orchestrator import execution_plugin
 from ..parser import validation
 from ..parser.consumption import ConsumptionContext
 from ..utils import collections, formatting, console
@@ -65,8 +67,8 @@ class ServiceBase(InstanceModelMixin):
     :vartype outputs: {basestring: :class:`Parameter`}
     :ivar workflows: Custom workflows that can be performed on the service
     :vartype workflows: {basestring: :class:`Operation`}
-    :ivar plugin_specifications: Plugins used by the service
-    :vartype plugin_specifications: {basestring: :class:`PluginSpecification`}
+    :ivar plugins: Plugins used by the service
+    :vartype plugins: {basestring: :class:`Plugin`}
     :ivar created_at: Creation timestamp
     :vartype created_at: :class:`datetime.datetime`
     :ivar updated_at: Update timestamp
@@ -176,13 +178,12 @@ class ServiceBase(InstanceModelMixin):
         return relationship.many_to_many(cls, 'parameter', prefix='outputs', dict_key='name')
 
     @declared_attr
-    def plugin_specifications(cls):
-        return relationship.many_to_many(cls, 'plugin_specification', dict_key='name')
+    def plugins(cls):
+        return relationship.many_to_many(cls, 'plugin', dict_key='name')
 
     # endregion
 
     description = Column(Text)
-
     created_at = Column(DateTime, nullable=False, index=True)
     updated_at = Column(DateTime)
 
@@ -207,6 +208,18 @@ class ServiceBase(InstanceModelMixin):
                 satisfied = False
         return satisfied
 
+    def find_hosts(self):
+        for node in self.nodes.itervalues():
+            node.find_host()
+
+    def configure_operations(self):
+        for node in self.nodes.itervalues():
+            node.configure_operations()
+        for group in self.groups.itervalues():
+            group.configure_operations()
+        for operation in self.workflows.itervalues():
+            operation.configure()
+
     def is_node_a_target(self, target_node):
         for node in self.nodes.itervalues():
             if self._is_node_a_target(node, target_node):
@@ -215,11 +228,11 @@ class ServiceBase(InstanceModelMixin):
 
     def _is_node_a_target(self, source_node, target_node):
         if source_node.outbound_relationships:
-            for the_relationship in source_node.outbound_relationships:
-                if the_relationship.target_node.name == target_node.name:
+            for relationship_model in source_node.outbound_relationships:
+                if relationship_model.target_node.name == target_node.name:
                     return True
                 else:
-                    node = the_relationship.target_node
+                    node = relationship_model.target_node
                     if node is not None:
                         if self._is_node_a_target(node, target_node):
                             return True
@@ -282,38 +295,25 @@ class ServiceBase(InstanceModelMixin):
             if not self.is_node_a_target(node):
                 self._dump_graph_node(node)
 
-    def _dump_graph_node(self, node):
+    def _dump_graph_node(self, node, capability=None):
         context = ConsumptionContext.get_thread_local()
         console.puts(context.style.node(node.name))
+        if capability is not None:
+            console.puts('{0} ({1})'.format(context.style.property(capability.name),
+                                            context.style.type(capability.type.name)))
         if node.outbound_relationships:
             with context.style.indent:
-                for the_relationship in node.outbound_relationships:
-                    relationship_name = context.style.property(the_relationship.name)
-                    if the_relationship.type is not None:
-                        relationship_type = context.style.type(the_relationship.type.name)
+                for relationship_model in node.outbound_relationships:
+                    relationship_name = context.style.property(relationship_model.name)
+                    if relationship_model.type is not None:
+                        console.puts('-> {0} ({1})'.format(relationship_name,
+                                                           context.style.type(
+                                                               relationship_model.type.name)))
                     else:
-                        relationship_type = None
-                    if the_relationship.target_capability is not None:
-                        capability_name = \
-                            context.style.node(the_relationship.target_capability.name)
-                    else:
-                        capability_name = None
-                    if capability_name is not None:
-                        if relationship_type is not None:
-                            console.puts('-> {0} ({1}) {2}'.format(relationship_name,
-                                                                   relationship_type,
-                                                                   capability_name))
-                        else:
-                            console.puts('-> {0} {1}'.format(relationship_name, capability_name))
-                    else:
-                        if relationship_type is not None:
-                            console.puts('-> {0} ({1})'.format(relationship_name,
-                                                               relationship_type))
-                        else:
-                            console.puts('-> {0}'.format(relationship_name))
-                    target_node = the_relationship.target_node
+                        console.puts('-> {0}'.format(relationship_name))
                     with console.indent(3):
-                        self._dump_graph_node(target_node)
+                        self._dump_graph_node(relationship_model.target_node,
+                                              relationship_model.target_capability)
 
 
 class NodeBase(InstanceModelMixin):
@@ -360,8 +360,10 @@ class NodeBase(InstanceModelMixin):
     :vartype policies: [:class:`Policy`]
     :ivar substitution_mapping: Our contribution to service substitution
     :vartype substitution_mapping: :class:`SubstitutionMapping`
-    :ivar tasks: Tasks on this node
+    :ivar tasks: Tasks for this node
     :vartype tasks: [:class:`Task`]
+    :ivar hosted_tasks: Tasks on this node
+    :vartype hosted_tasks: [:class:`Task`]
     """
 
     __tablename__ = 'node'
@@ -417,7 +419,7 @@ class NodeBase(InstanceModelMixin):
 
     @property
     def is_available(self):
-        return self.state not in [self.INITIAL, self.DELETED, self.ERROR]
+        return self.state not in (self.INITIAL, self.DELETED, self.ERROR)
 
     # region foreign_keys
 
@@ -455,7 +457,7 @@ class NodeBase(InstanceModelMixin):
     # region one_to_one relationships
 
     @declared_attr
-    def host(cls):
+    def host(cls): # pylint: disable=method-hidden
         return relationship.one_to_one_self(cls, 'host_fk')
 
     # endregion
@@ -522,17 +524,9 @@ class NodeBase(InstanceModelMixin):
     __mapper_args__ = {'version_id_col': version} # Enable SQLAlchemy automatic version counting
 
     @property
-    def ip(self):
-        # TODO: totally broken
-        if not self.host_fk:
-            return None
-        host_node = self.host
-        if 'ip' in host_node.runtime_properties:  # pylint: disable=no-member
-            return host_node.runtime_properties['ip']  # pylint: disable=no-member
-        host_node = host_node.node_template  # pylint: disable=no-member
-        host_ip_property = host_node.properties.get('ip')
-        if host_ip_property:
-            return host_ip_property.value
+    def host_address(self):
+        if self.host and self.host.runtime_properties:
+            return self.host.runtime_properties.get('ip')
         return None
 
     def satisfy_requirements(self):
@@ -567,9 +561,10 @@ class NodeBase(InstanceModelMixin):
             if target_node_capability is not None:
                 # Relate to the first target node that has capacity
                 for node in target_nodes:
-                    target_capability = node.capabilities.get(target_node_capability.name)
-                    if target_capability.relate():
+                    a_target_capability = node.capabilities.get(target_node_capability.name)
+                    if a_target_capability.relate():
                         target_node = node
+                        target_capability = a_target_capability
                         break
             else:
                 # Use first target node
@@ -577,14 +572,15 @@ class NodeBase(InstanceModelMixin):
 
             if target_node is not None:
                 if requirement_template.relationship_template is not None:
-                    the_relationship = \
+                    relationship_model = \
                         requirement_template.relationship_template.instantiate(self)
                 else:
-                    the_relationship = models.Relationship(target_capability=target_capability)
-                the_relationship.name = requirement_template.name
-                the_relationship.requirement_template = requirement_template
-                the_relationship.target_node = target_node
-                self.outbound_relationships.append(the_relationship)
+                    relationship_model = models.Relationship()
+                relationship_model.name = requirement_template.name
+                relationship_model.requirement_template = requirement_template
+                relationship_model.target_node = target_node
+                relationship_model.target_capability = target_capability
+                self.outbound_relationships.append(relationship_model)
                 return True
             else:
                 context.validation.report('requirement "{0}" of node "{1}" targets node '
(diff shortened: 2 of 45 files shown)


aria/modeling/orchestration.py
```diff
@@ -200,15 +200,52 @@ class PluginBase(ModelMixin):
...
+    :ivar retry_count: How many retries occurred
+    :vartype retry_count: int
     """
 
     __tablename__ = 'task'
 
     __private_fields__ = ['node_fk',
                           'relationship_fk',
                           'plugin_fk',
-                          'execution_fk',
```
ran-z: ?

aria/modeling/orchestration.py
```diff
@@ -258,29 +290,17 @@ def execution(cls):
...
-
-    due_at = Column(DateTime, nullable=False, index=True, default=datetime.utcnow())
-    started_at = Column(DateTime, default=None)
-    ended_at = Column(DateTime, default=None)
+    implementation = Column(String)
     max_attempts = Column(Integer, default=1)
-    retry_count = Column(Integer, default=0)
     retry_interval = Column(Float, default=0)
     ignore_failure = Column(Boolean, default=False)
+    due_at = Column(DateTime, nullable=False, index=True, default=datetime.utcnow())
```
ran-z: if you already do the separation to "state" fields and other, this should probably be under state.

aria/orchestrator/workflows/api/task.py
```diff
@@ -65,86 +65,82 @@ def __init__(self,
...
         """
 
-        assert isinstance(actor, (models.Node, models.Relationship))
-        assert actor_type in ('node', 'relationship')
         assert interface_name and operation_name
-        assert runs_on in models.Task.RUNS_ON
         super(OperationTask, self).__init__()
 
         self.actor = actor
+        self.actor_type = actor_type
```
ran-z: why are any changes needed here (besides runs_on)?

extensions/aria_extension_tosca/profiles/tosca-simple-1.0/interfaces.yaml
```diff
@@ -63,24 +63,48 @@ interface_types:
     pre_configure_source:
       description: >-
         Operation to pre-configure the source endpoint.
+      _extensions:
+        default_dependencies:
+          - edge > source
```
ran-z: as we mentioned, these should not be a part of the dependencies. it should sit elsewhere.

ran-z: furthermore, for the time being, it's not "default", but rather the only place where one can determine where the operation shall run.

tests/modeling/test_models.py
```diff
@@ -583,48 +583,49 @@ def test_node_model_creation(self, node_template_storage, is_valid, name, runtim
                    node_template_storage.service.list()[0]
 
 
-class TestNodeIP(object):
+class TestNodeHostAddress(object):
```
ran-z: why change this everywhere?

tests/resources/service-templates/tosca-simple-1.0/node-cellar/node-cellar.yaml
```diff
@@ -155,14 +155,34 @@ topology_template:
...
                       - max_instances: { greater_or_equal: 8 }
-    
+            relationship:
+              interfaces:
+                Configure:
+                  target_changed:
+                    implementation:
+                      primary: changed.sh
+                      dependencies:
+                        - edge > target
```
ran-z: remove

aria/parser/consumption/modeling.py
```diff
@@ -145,6 +145,24 @@ def consume(self):
         self.context.modeling.instance.validate_capabilities()
 
 
+class FindHosts(Consumer):
```
ran-z: why is this implemented as another consumer?

aria/orchestrator/execution_plugin/instantiation.py
```diff
@@ -0,0 +1,188 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ...utils.formatting import full_type_name
+from ...utils.collections import OrderedDict
+from ...parser import validation
+from ...parser.consumption import ConsumptionContext
+
+
+def configure_operation(operation):
```
ran-z: almost everything in this module needs to go. this method is okay, the rest should be transparent to the execution plugin

aria/modeling/service_instance.py
```diff
@@ -1407,6 +1441,19 @@ def operation_template_fk(cls):
 
     # endregion
 
+    def configure(self):
```
ran-z: this definitely cant be here.

aria/modeling/service_instance.py
```diff
@@ -180,6 +181,16 @@ def validate_capabilities(self):
                 satisfied = False
         return satisfied
 
+    def find_hosts(self):
```
ran-z: more methods on the models which we'll want to remove in a short while :(
i mean, this method isn't relevant for an instantiated service - it's only used for its (and its dependencies) creation. why would it sit as part of the instance?

aria/modeling/service_instance.py
```diff
@@ -529,6 +524,32 @@ def validate_capabilities(self):
...
+                        return host
+            for the_relationship in node.inbound_relationships:
+                if (the_relationship.target_capability is not None) and \
+                    the_relationship.target_capability.type.role == 'feature':
+                    host = _find_host(the_relationship.source_node)
+                    if host is not None:
+                        return host
+            return None
+
+        self.host = _find_host(self)
```
ran-z: why is this dynamic? are we expecting this to change for an existing node?

tests/modeling/test_models.py
mxmrlv: no new tests were written for the new feature.

aria/modeling/service_instance.py
tliron: That's not true: this phase will need be run every time there is a change in the topology. This is part of the reason I don't think `instantiation` is the right name for this module we talk about, since it would actually be used in many cases after instantiation.

aria/modeling/orchestration.py
tliron: The diff algorithm borked here, if you look at the file it makes sense.

aria/modeling/orchestration.py
tliron: Does `due_at` ever change after creation? I guess you're implying that it does...

aria/modeling/service_instance.py
tliron: OK, so where? We don't have these separate module yet and I did not think I should be creating it as part of this PR.

aria/orchestrator/execution_plugin/instantiation.py
tliron: OK, where do you suggest moving them?

aria/orchestrator/workflows/api/task.py
tliron: Because we decided together that the logic that was here should not be in operation creation, but in service instance creation.

aria/parser/consumption/modeling.py
tliron: This is just how phases work right now in "the parser". We keep talking about all of this moving into a separate "instantiation" module, but this is what we have for now.

extensions/aria_extension_tosca/profiles/tosca-simple-1.0/interfaces.yaml
tliron: I thought you said that if there is no cost to allowing the user to override it, we could.

tests/modeling/test_models.py
tliron: That whole part of the code (finding the "ip") had to be refactored because it was essentially broken. As I was changing it anyway it made sense to change the name. It's really not hard to change back.

aria/modeling/service_instance.py
```diff
@@ -529,6 +524,32 @@ def validate_capabilities(self):
                 satisfied = False
         return satisfied
 
+    def find_host(self):
+        def _find_host(node):
```
tliron: If we do that then we'll be calling `find_host` again on again on the same nodes that are targets of different relationships. So, to do that efficiently we would need some extra flag (`found_host`?) to signify that we don't have to do that again for that node. It made more sense to me to have it set once and only once per node.

aria/modeling/service_instance.py
```diff
@@ -529,6 +524,32 @@ def validate_capabilities(self):
...
+            if node.type.role == 'host':
+                return node
+            for the_relationship in node.outbound_relationships:
+                if (the_relationship.target_capability is not None) and \
+                    the_relationship.target_capability.type.role == 'host':
+                    host = _find_host(the_relationship.target_node)
+                    if host is not None:
+                        return host
+            for the_relationship in node.inbound_relationships:
+                if (the_relationship.target_capability is not None) and \
```
tliron: This is a bit hard to explain. We tend to think of the host relationship as "contained in", but TOSCA also allows for "hosts a" kind of relationship, for example for features. See [here](https://github.com/apache/incubator-ariatosca/blob/ARIA-92-plugin-in-implementation-string/tests/resources/service-templates/tosca-simple-1.0/node-cellar/node-cellar.yaml#L177): to find out the loadbalancer node's host, we need to look inbound that it is a feature of Nginx.

So, bottom line is that this function has to be bi-recursive, both inbound and outbound. Yay. :)

aria/modeling/service_instance.py
tliron: I don't know what you mean by "dynamic" here. These functions must be called during the initial instantiation phase, and then will need to be called again whenever there is a change in the topology.

aria/modeling/service_template.py
```diff
@@ -1544,11 +1547,24 @@ def as_raw(self):
...
         from . import models
+        from ..orchestrator import context
+        plugin = None
+        if self.plugin_specification is not None:
+            try:
+                workflow_context = context.workflow.current.get()
+                plugin = self.plugin_specification.find_plugin(workflow_context.model.plugin.list())
+            except context.exceptions.ContextException:
+                pass
+                # TODO
```
tliron: Yes, we need to talk about this. :) I don't know how else to find other models. Is there a current context for SQLAlchemy?

aria/modeling/orchestration.py
ran-z: yup :)

aria/modeling/service_instance.py
```diff
@@ -522,17 +524,10 @@ def properties(cls):
...
-            return None
-        host_node = self.host
-        if 'ip' in host_node.runtime_properties:  # pylint: disable=no-member
-            return host_node.runtime_properties['ip']  # pylint: disable=no-member
-        host_node = host_node.node_template  # pylint: disable=no-member
-        host_ip_property = host_node.properties.get('ip')
-        if host_ip_property:
-            return host_ip_property.value
+    def host_address(self):
+        if self.host:
```
ran-z: &&? :P

aria/modeling/service_instance.py
```diff
@@ -522,17 +524,10 @@ def properties(cls):
...
+            if self.host.runtime_properties:
+                return self.host.runtime_properties.get('host_address')
```
ran-z: while I understand your reasoning to using "host_address" instead of "ip" everywhere, the actual runtime property that gets set by plugins remains to be "ip". this can't change at this stage.

aria/modeling/service_instance.py
ran-z: :+1:

aria/modeling/service_instance.py
ran-z: im still a bit confused by this myself - caching the host is definitely a good idea, but it looks here like the internal method _find_host is called recursively repeatedly.
I'm not sure this is exactly the same as what Maxim meant, but I think I'd expect each node's host finding method to run once at most, so that would mean either "host" is a self-initializing property, or a field initialized at its constructor, or you could use the "find_host" method, but basically what i'm missing here is the check for whether the value has been cached already or not.

i'd expect it to be more like:
```
def find_host(self):
    def _find_host(node):
    ....
        host = the_relationship.target_node.find_host()
    ....

if not self.host:  # or indeed use some extra flag
    self.host = _find_host(node)    
return self.host
```

aria/modeling/service_instance.py
```diff
@@ -1632,9 +1673,9 @@ def interface_fk(cls):
         return relationship.foreign_key('interface', nullable=True)
 
     @declared_attr
-    def plugin_specification_fk(cls):
-        """For Operation one-to-one to PluginSpecification"""
-        return relationship.foreign_key('plugin_specification', nullable=True)
+    def plugin_fk(cls):
```
ran-z: does it make sense for operation no longer to have an fk for plugin specification at all?
is that going to remain for operation template?

aria/modeling/service_instance.py
```diff
@@ -1685,12 +1725,31 @@ def inputs(cls):
...
     implementation = Column(Text)
+    configuration = Column(modeling_types.StrictDict(key_cls=basestring))
     dependencies = Column(modeling_types.StrictList(item_cls=basestring))
     executor = Column(Text)
     max_retries = Column(Integer)
     retry_interval = Column(Integer)
 
+    def configure(self):
+        from . import models
+        if (self.implementation is None) or (self.interface is None):
```
ran-z: how can interface be None here?

aria/modeling/service_instance.py
```diff
@@ -1685,12 +1725,31 @@ def inputs(cls):
...
+            return
+
+        if self.plugin is None:
+            arguments = configure_operation(self)
```
ran-z: please use `execution_plugin.instantiation.configure_operation` or so, there are enough `configure_operation` calls in this module as it is and it's easily confusing :)

tests/resources/service-templates/tosca-simple-1.0/node-cellar/node-cellar.yaml
```diff
@@ -98,8 +98,8 @@ topology_template:
               #token: { get_property: [ HOST, flavor_name ] }
       interfaces:
         Maintenance:
-          enable: juju > charm.maintenance_on
-          disable: juju > charm.maintenance_off
+          enable: juju \> charm.maintenance_on
```
ran-z: wait so what's this symbol about again? :)
Why are we not using the same symbol as we do for `dependencies`?

extensions/aria_extension_tosca/simple_v1_0/modeling/capabilities.py
```diff
@@ -78,6 +78,11 @@ def get_inherited_capability_definitions(context, presentation, for_presentation
                 #capability_definitions[capability_name] = capability_definition
             else:
                 capability_definition = our_capability_definition._clone(for_presentation)
+                if isinstance(capability_definition._raw, basestring):
+                    # # Make sure we have a dict
```
ran-z: double comment

aria/orchestrator/workflows/builtin/utils.py
```diff
@@ -71,15 +71,13 @@ def relationship_tasks(
...
-                                           runs_on='source')
+                                           operation_name=source_operation_name)
         )
     if target_operation_name:
         operations.append(
             OperationTask.for_relationship(relationship=relationship,
                                            interface_name=interface_name,
-                                           operation_name=target_operation_name,
-                                           runs_on='target')
+                                           operation_name=target_operation_name,)
```
ran-z: tuple?

aria/parser/consumption/modeling.py
ran-z: :+1:

tests/modeling/test_models.py
ran-z: :+1:

aria/modeling/orchestration.py
ran-z: :+1:

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
```diff
@@ -352,20 +359,44 @@ def create_interface_template_model(context, service_template, interface):
...
+            model.plugin_specification = service_template.plugin_specifications.get(plugin_name)
+            if model.plugin_specification is None:
+                context.validation.report(
+                    'unknown plugin "%s" specified in operation implementation: %s'
+                    % (plugin_name, primary),
+                    locator=operation._get_child_locator('implementation', 'primary'),
+                    level=Issue.BETWEEN_TYPES)
+    
+        relationship_edge = operation._get_extensions(context).get('relationship_edge')
+        if relationship_edge is not None:
```
ran-z: what if it is None? where do we set the default? I think either we don't or I've missed it :)

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
```diff
@@ -352,20 +359,44 @@ def create_interface_template_model(context, service_template, interface):
...
-            parse_implementation_string(context, service_template, operation.implementation.primary)
-
+    if implementation is not None: 
+        primary = implementation.primary
+        plugin_name, model.implementation = split_prefix(primary)
+        if plugin_name is not None:
+            model.plugin_specification = service_template.plugin_specifications.get(plugin_name)
+            if model.plugin_specification is None:
+                context.validation.report(
+                    'unknown plugin "%s" specified in operation implementation: %s'
```
ran-z: is this message clear enough about the fact that the plugin is not properly declared as a policy in the service template (rather than the question of whether it exists or not in the plugin repository)?

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
```diff
@@ -461,8 +492,15 @@ def create_workflow_operation_template_model(context, service_template, policy):
     properties = policy._get_property_values(context)
     for prop_name, prop in properties.iteritems():
         if prop_name == 'implementation':
-            model.plugin_specification, model.implementation = \
-                parse_implementation_string(context, service_template, prop.value)
+            plugin_name, model.implementation = split_prefix(prop.value)
+            if plugin_name is not None:
```
ran-z: duplicate code, why not refactor this?

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
```diff
@@ -677,21 +715,18 @@ def pattern(node_type, container): # pylint: disable=unused-argument
...
-    if plugin_name == 'execution':
-        plugin_specification = None
-    else:
-        plugin_specification = service_template.plugin_specifications.get(plugin_name)
-        if plugin_specification is None:
-            raise ValueError('unknown plugin: "{0}"'.format(plugin_name))
 
-    implementation = implementation[index+1:].strip()
-    return plugin_specification, implementation
+def set_nested(the_dict, keys, value):
```
ran-z: could you add a few comments here, possibly an example of input --> output, just so it's clearer? thanks

aria/modeling/service_instance.py
ran-z: :+1:

aria/modeling/service_instance.py
ran-z: :+1:

aria/modeling/service_template.py
```diff
@@ -1848,11 +1856,30 @@ def as_raw(self):
...
     def instantiate(self, container):
         from . import models
+        from ..orchestrator import context
+        plugin = None
+        if self.plugin_specification is not None:
+            # TODO: we are planning a separate "instantiation" module where this will be called or
+            # moved to. There, we will probably have a context with a storage manager. Until then,
+            # this is the only potentially available context, which of course will only be available
+            # if we're in a workflow.
+            plugin = None
```
ran-z: redundant

aria/modeling/service_template.py
```diff
@@ -1848,11 +1856,30 @@ def as_raw(self):
 
     def instantiate(self, container):
         from . import models
+        from ..orchestrator import context
+        plugin = None
+        if self.plugin_specification is not None:
```
ran-z: I'm not sure I understand this section.
Operations are instantiated at service creation time. No workflow context should be available at this time.
If a storage model object is required, it should be passed into the instantiate module directly (the current interface is problematic as is, and I've already broken it on the CLI branch in order to pass inputs to service template's instantiate), not taken off some thread local context.

aria/orchestrator/workflows/api/task.py
```diff
@@ -66,98 +66,84 @@ def __init__(self,
...
         assert interface_name and operation_name
-        assert runs_on in models.Task.RUNS_ON
         super(OperationTask, self).__init__()
 
+        operation = None
+        interface = actor.interfaces.get(interface_name)
+        if interface is not None:
+            operation = interface.operations.get(operation_name)
+
+        if operation is None:
```
ran-z: (FYI this section is refactored elsewhere in the CLI branch, so it will get overridden here too)

aria/orchestrator/workflows/api/task.py
```diff
@@ -66,98 +66,84 @@ def __init__(self,
...
+            raise exceptions.OperationNotFoundException(
+                'Could not find operation "{0}" on interface "{1}" for {2} "{3}"'
+                .format(operation_name, interface_name, actor_type, actor.name))
+
+        if operation.implementation is None:
```
ran-z: Why should an empty operation raise an error (rather than do nothing)?
if a node template has its type as root, and implements `create` with a proper implementation but does not use `configure`, would it not appear as an empty implementation operation?

aria/modeling/service_instance.py
tliron: Correct.

aria/modeling/service_instance.py
tliron: It will be None if the operation is attached to `Service` as a workflow :) I know you suggested a different model entirely for workflows ... that would be a separate JIRA perhaps.

aria/modeling/service_instance.py
tliron: :+1:

tests/resources/service-templates/tosca-simple-1.0/node-cellar/node-cellar.yaml
tliron: I've put the backslash to escape it and disable it for now. :/ Since the Juju plugin is not installed, the tests that use this would fail otherwise.

aria/orchestrator/execution_plugin/instantiation.py
```diff
@@ -0,0 +1,187 @@
...
+    from . import operations
+    operation.implementation = '{0}.{1}'.format(operations.__name__,
+                                                operations.run_script_locally.__name__)
+
+
+def _configure_remote(operation, configuration, arguments):
+    """
+    Remote SSH operation via Fabric.
+    """
+    default_user = 'admin'
```
ran-z: why these defaults? wouldnt it be better to fail?

extensions/aria_extension_tosca/simple_v1_0/modeling/capabilities.py
tliron: :+1:

aria/orchestrator/execution_plugin/instantiation.py
```diff
@@ -0,0 +1,187 @@
...
+    arguments['use_sudo'] = ssh.get('use_sudo')
+    arguments['hide_output'] = ssh.get('hide_output')
+    arguments['fabric_env'] = {}
+    if 'warn_only' in ssh:
+        arguments['fabric_env']['warn_only'] = ssh['warn_only']
+    arguments['fabric_env']['user'] = ssh.get('user')
+    arguments['fabric_env']['password'] = ssh.get('password')
+    arguments['fabric_env']['key'] = ssh.get('key')
+    arguments['fabric_env']['key_filename'] = ssh.get('key_filename')
+    if 'address' in ssh:
```
ran-z: i assume you came up with this one?

aria/orchestrator/execution_plugin/instantiation.py
```diff
@@ -0,0 +1,187 @@
...
+    Remote SSH operation via Fabric.
+    """
+    default_user = 'admin'
+    default_password = 'admin'
+    ssh = _get_ssh(configuration.pop('ssh')) if 'ssh' in configuration else {}
+    if 'user' not in ssh:
+        ssh['user'] = default_user
+    if ('password' not in ssh) and ('key' not in ssh) and ('key_filename' not in ssh):
+        ssh['password'] = default_password
+    arguments['use_sudo'] = ssh.get('use_sudo')
```
ran-z: why not use copy of some sort?

aria/orchestrator/workflows/builtin/utils.py
tliron: :+1: (though it's totally legally in Python)

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
tliron: None is the default -- it means that nowhere did the user set the edge explicitly set one way or the other (the column is nullable).

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
tliron: :+1:

aria/orchestrator/execution_plugin/instantiation.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+def _validate_type(value, the_type, name):
+    if not isinstance(value, the_type):
+        context = ConsumptionContext.get_thread_local()
+        context.validation.report('"{0}" configuration is not a {1}'
+                                  .format(name, full_type_name(the_type)),
+                                  level=validation.Issue.BETWEEN_TYPES)
+
+def _str_to_bool(value, name):
```
ran-z: this and the next method need to sit elsewhere. on CLI branch there's a `type` module under `utils`, and the `dict_to_list`  should go to where there's already similar methods.

the validation thats called within these methods can be called before or after them

aria/orchestrator/execution_plugin/instantiation.py
```diff
@@ -0,0 +1,187 @@
...
+                                      level=validation.Issue.BETWEEN_TYPES)
+    return value
+
+
+def _get_ssh(value):
+    if value is None:
+        return {}
+    _validate_type(value, dict, 'ssh')
+    for k, v in value.iteritems():
+        if k == 'use_sudo':
```
ran-z: i think it'd probably have been neater to have a dict mapping field-name to type and then validating it rather than having to add here manually every new field that might get introduced

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
tliron: :+1:

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
tliron: :+1:

aria/modeling/service_template.py
tliron: :+1:

aria/modeling/service_template.py
tliron: You are correct, I had a long conversation with Maxim about it. Eventually when we have an instantiation module, it is very likely that the module will have a context. See comment at line 1861.

aria/orchestrator/workflows/api/task.py
tliron: What is the meaning of an empty implementation operation? The operation would fail in the orchestrator, seems much better to catch that here and not create that operation.

aria/orchestrator/execution_plugin/instantiation.py
tliron: Oops, leftover from devs. :+1:

aria/orchestrator/execution_plugin/instantiation.py
tliron: Well, I read our code and figured these out. :) They were not well documented anywhere.

aria/orchestrator/execution_plugin/instantiation.py
tliron: Due to the various validations. The structure is also a little bit different. In the service template, the user doesn't have to write "fabric_env" (or really shouldn't care about our implementation detail at all), the just write "ssh.password".

aria/orchestrator/execution_plugin/instantiation.py
tliron: I thought it would be more straightforward to do this than have a generic mechanism. Since *only* this plugin does this (for all the rest the configuration params get merged into inputs) it seemed unnecessary for me to create a generic mechanism.

aria/modeling/service_template.py
ran-z: I understand and I have seen the comment, but I'm saying the changes I've mentioned need to happen now, not after we have the instantiation module.
More so, I don't understand how this is expected to work anyway, as no execution context is expect to be present at the time this method is called.

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
```diff
@@ -677,21 +698,47 @@ def pattern(node_type, container): # pylint: disable=unused-argument
...
+        set_nested(the_dict[key], keys, value)
+
+
+def parse_implementation_string(context, service_template, presentation, model, implementation):
+    plugin_name, model.implementation = split_prefix(implementation)
+    if plugin_name is not None:
+        model.plugin_specification = service_template.plugin_specifications.get(plugin_name)
+        if model.plugin_specification is None:
+            context.validation.report(
+                'no policy for plugin "%s" specified in operation implementation: %s'
```
ran-z: use string format please :sweat_smile:

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
ran-z: Don't you think it'd be better to have it be set to `source` by default?
I assume when it's `None` we should treat it as `source`, so I think it'd be better if the data model reflects this design decision, no?

aria/orchestrator/workflows/builtin/utils.py
ran-z: yes i know, I just figured this is not what you wanted to have there :sweat_smile: anyway i see its fixed, thanks

tests/resources/service-templates/tosca-simple-1.0/node-cellar/node-cellar.yaml
ran-z: :+1:

aria/modeling/service_instance.py
ran-z: oh, right. Please add a tiny comment next to it explaining that this case is only relevant for "Workflow operations" because I'm afraid I'll wonder about it the next time I see it as well :sweat_smile:  thanks!

aria/modeling/service_instance.py
ran-z: :+1:

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
tliron: :+1:

aria/orchestrator/execution_plugin/instantiation.py
ran-z: :+1:

aria/orchestrator/execution_plugin/instantiation.py
ran-z: :+1:

aria/modeling/service_instance.py
ran-z: :+1:

aria/modeling/service_instance.py
tliron: :+1:

aria/modeling/service_instance.py
tliron: :+1:

tests/resources/service-templates/tosca-simple-1.0/node-cellar/node-cellar.yaml
tliron: I'm changing this back. :) Will explain the new behavior on Slack.

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
tliron: We agreed that it's better to keep this "edge was not set" information.

aria/orchestrator/workflows/api/task.py
tliron: We agreed to keep this behavior for now.

aria/modeling/service_template.py
```diff
@@ -1864,13 +1865,21 @@ def as_raw(self):
 
     def instantiate(self, container):
         from . import models
-        plugin = self.plugin_specification.find_plugin() \
-            if self.plugin_specification is not None else None
+        if self.plugin_specification and self.plugin_specification.enabled:
+            plugin = self.plugin_specification.plugin
+            implementation = self.implementation if plugin is not None else None
+            # "plugin" would be none if a match was not found. In that case, a validation error
```
ran-z: im not sure i understood this comment.
i think you're trying to say that `plugin` can't be None here since the test for enabled has already been performed earlier?
the first line of the comment makes it sound like `plugin` can be None when it cant

aria/modeling/service_template.py
```diff
@@ -2077,27 +2091,39 @@ def service_template_fk(cls):
...
             ('name', self.name),
-            ('version', self.version)))
+            ('version', self.version),
+            ('enabled', self.enabled)))
 
     def coerce_values(self, container, report_issues):
         pass
 
     def instantiate(self, container):
-        from . import models
```
ran-z: wait, so instantiate is now empty? :confused:

PR #96, opened 2017-04-05
ARIA-137-Support-for-predicate-based-queries-in-the-SQL-mapi

mxmrlv:


aria/storage/sql_mapi.py
```diff
@@ -269,12 +279,24 @@ def _get_query(self,
...
         query = self._filter_query(query, filters)
         query = self._sort_query(query, sort)
         return query
 
+    @staticmethod
+    def _convert_operands(filters):
+        for column, conditions in filters.items():
+            if isinstance(conditions, dict):
+                for predicate, operand in conditions.items():
+                    if predicate in _predicates:
```
ran-z: else raise maybe?

aria/storage/sql_mapi.py
```diff
@@ -243,7 +250,10 @@ def _filter_query(self, query, filters):
     @staticmethod
     def _add_value_filter(query, filters):
         for column, value in filters.items():
-            if isinstance(value, (list, tuple)):
+            if isinstance(value, dict):
+                for predicate, operand in value.items():
+                    query = query.filter(getattr(column, predicate)(operand))
```
ran-z: hmm is this really how you do this? :D seems complex

mxmrlv: well, SQL makes it fairly simple. as in `value > other_value`. but propagating it to the query is messy...

PR #100, opened 2017-04-14
ARIA-140 Version utils

tliron:


aria/utils/versions.py
```diff
@@ -0,0 +1,162 @@
...
+                return _NULL
+            qualifier = value
+    else:
+        # Version strings with no qualifiers are higher
+        qualifier = _INF
+
+    return primary, qualifier
+
+
+VersionString.NULL = VersionString()
```
AviaE: What is the purpose of this NULL? I don't see any use of it except in the tests.

aria/modeling/service_template.py
```diff
@@ -33,7 +33,8 @@
 from ..parser import validation
```
AviaE: Please expand the commit message by adding information on the purpose of your change/addition.

aria/utils/versions.py
```diff
@@ -0,0 +1,162 @@
...
+"""
+General-purpose version string handling
+"""
+
+import re
+
+
+_INF = float('inf')
+
+_NULL = (), _INF
```
tliron: Because "Numeric qualifiers will always be greater than prefixed integer qualifiers, e.g. "1.1-1" > "1.1-beta1". The qualifier is a float that is an extra, but if the qualifier is not present "infinity" should always win over it. It shouldn't matter what's "more natural" seeming: this is an underscored internal var that's never seen by users of the lib.

aria/utils/versions.py
```diff
@@ -0,0 +1,162 @@
...
+def parse_version_string(version): # pylint: disable=too-many-branches
+    """
+    Parses a version string.
+
+    :param version: The version string
+    :returns: The primary tuple and qualifier float
+    :rtype: ((int), float)
+    """
+
+    if version is None:
```
tliron: Already in line #64 do you see the default to be None. But anyway, this is a user-exposed function and None values do occur in the wild.

aria/utils/versions.py
tliron: Because "Numeric qualifiers will always be greater than prefixed integer qualifiers, e.g. "1.1-1" > "1.1-beta1". The qualifier is a float that is an extra, but if the qualifier is not present "infinity" should always win over it. It shouldn't matter what's "more natural" seeming: this is an underscored internal var that's never seen by users of the lib. Users instead use `VersionString.NULL` or can init `VersionString` with None.

aria/utils/versions.py
```diff
@@ -0,0 +1,162 @@
...
+        return _NULL
+    version = unicode(version)
```
tliron: It is true, there is no test for this function specifically with a None value. I will add it.

aria/utils/versions.py
```diff
@@ -0,0 +1,162 @@
...
+
+    # Split to primary and qualifier on '-'
+    split = version.split('-', 2)
+    if len(split) == 2:
+        primary, qualifier = split
+    else:
```
tliron: Oops, yes, changed to rsplit on 1.

aria/utils/versions.py
```diff
@@ -0,0 +1,162 @@
...
+        qualifier = None
+
+    # Parse primary
+    split = primary.split('.')
+    primary = []
+    for element in split:
+        if _DIGITS_RE.match(element) is None:
+            # Invalid version string
+            return _NULL
+        try:
```
tliron: Maybe it's too many digits for the current Python platform's size of int? Seems to me that it's best not to assume it would always work. Utility methods need to be especially robust.

aria/utils/versions.py
tliron: It's a constant for the user to use in any situation where they need a `NULL` value. It's customary to provide singletons for efficiency.

aria/modeling/service_template.py
tliron: I don't understand. What does this have to do with this line in the code? Where in the commit message? We have limited our commit messages to one line with reference to the JIRA for more information. I know you tend to add more info, but the rest of us don't normally.

aria/modeling/service_template.py
```diff
@@ -33,7 +33,8 @@
...
 from ..parser.consumption import ConsumptionContext
 from ..parser.reading import deepcopy_with_locators
-from ..utils import collections, formatting, console
+from ..utils import (collections, formatting, console)
```
tliron: We have discussed this and have agreed that it can be in one line if it fits. We have a lot of code that is like this.

aria/modeling/service_template.py
```diff
@@ -2131,13 +2132,15 @@ def resolve(self, model_storage):
...
         plugins = model_storage.plugin.list()
         matching_plugins = []
-        for plugin in plugins:
-            # TODO: we need to use a version comparator
-            if (plugin.name == self.name) and \
-                    ((self.version is None) or (plugin.package_version >= self.version)):
-                matching_plugins.append(plugin)
+        if plugins:
+            for plugin in plugins:
+                if (plugin.name == self.name) and \
```
tliron: I strongly disagree, especially in this case when there is `or` and `and` in the same clause. Do you otherwise expect the reader to know the priority of logical binary operations? Parentheses remove any potential confusion.

tliron:
Responses.

aria/utils/versions.py
tliron: This is a user-exposed function and None values do occur in the wild.

aria/utils/versions.py
tliron: It is true, there is no tests for this function specifically with a None value. I will add it.

aria/utils/versions.py
tliron: Because "Numeric qualifiers will always be greater than prefixed integer qualifiers, e.g. "1.1-1" > "1.1-beta1". The qualifier is a float that is an extra, but if the qualifier is not present "infinity" should always win over it. It shouldn't matter what's "more natural" seeming: this is an underscored internal var that's never seen by users of the lib. Users instead use `VersionString.NULL` or can init `VersionString` with None.

AviaE:
@tliron and I discussed his replies via zoom, and this pull request is free to merge (with minor changes).

PR #107, opened 2017-04-20
ARIA-148 CLI display commands

tliron:
Also includes some typo fixes for the CLI.

### DIFF
--- a/aria/cli/commands/service_templates.py
+++ b/aria/cli/commands/service_templates.py
@@ -23,11 +23,13 @@ from .. import utils
 from ..core import aria
 from ...core import Core
 from ...storage import exceptions as storage_exceptions
+from ...parser import consumption
+from ...utils import (formatting, collections, console)
 
 
 DESCRIPTION_FIELD_LENGTH_LIMIT = 20
 SERVICE_TEMPLATE_COLUMNS = \
-    ['id', 'name', 'description', 'main_file_name', 'created_at', 'updated_at']
+    ('id', 'name', 'description', 'main_file_name', 'created_at', 'updated_at')
 
 
 @aria.group(name='service-templates')
@@ -43,32 +45,52 @@ def service_templates():
 @aria.argument('service-template-name')
 @aria.options.verbose()
 @aria.pass_model_storage
+@aria.options.service_template_mode_full
+@aria.options.mode_types
+@aria.options.format_json
+@aria.options.format_yaml
 @aria.pass_logger
-def show(service_template_name, model_storage, logger):
-    """Show information for a specific service templates
+def show(service_template_name, model_storage, mode_full, mode_types, format_json, format_yaml,
+         logger):
+    """Show information for a specific service template
 
     `SERVICE_TEMPLATE_NAME` is the name of the service template to show information on.
     """
-    logger.info('Showing service template {0}...'.format(service_template_name))
     service_template = model_storage.service_template.get_by_name(service_template_name)
-    service_template_dict = service_template.to_dict()
-    service_template_dict['#services'] = len(service_template.services)
 
-    column_formatters = \
-        dict(description=table.trim_formatter_generator(DESCRIPTION_FIELD_LENGTH_LIMIT))
-    columns = SERVICE_TEMPLATE_COLUMNS + ['#services']
-    table.print_data(columns, service_template_dict, 'Service-template:',
-                     column_formatters=column_formatters, col_max_width=50)
+    if format_json or format_yaml:
+        mode_full = True
 
-    if service_template_dict['description'] is not None:
-        logger.info('Description:')
-        logger.info('{0}{1}'.format(service_template_dict['description'].encode('UTF-8') or '',
-                                    os.linesep))
+    if mode_full:
+        consumption.ConsumptionContext()
+        if format_json:
+            console.puts(formatting.json_dumps(collections.prune(service_template.as_raw)))
+        elif format_yaml:
+            console.puts(formatting.yaml_dumps(collections.prune(service_template.as_raw)))
+        else:
+            service_template.dump()
+    elif mode_types:
+        consumption.ConsumptionContext()
+        service_template.dump_types()
+    else:
+        logger.info('Showing service template {0}...'.format(service_template_name))
+        service_template_dict = service_template.to_dict()
+        service_template_dict['#services'] = len(service_template.services)
+        columns = SERVICE_TEMPLATE_COLUMNS + ('#services',)
+        column_formatters = \
+            dict(description=table.trim_formatter_generator(DESCRIPTION_FIELD_LENGTH_LIMIT))
+        table.print_data(columns, service_template_dict, 'Service-template:',
+                         column_formatters=column_formatters, col_max_width=50)
 
-    if service_template.services:
-        logger.info('Existing services:')
-        for service in service_template.services:
-            logger.info('\t{0}'.format(service.name))
+        if service_template_dict['description'] is not None:
+            logger.info('Description:')
+            logger.info('{0}{1}'.format(service_template_dict['description'].encode('UTF-8') or '',
+                                        os.linesep))
+
+        if service_template.services:
+            logger.info('Existing services:')
+            for service in service_template.services:
+                logger.info('\t{0}'.format(service.name))
 
 
 @service_templates.command(name='list',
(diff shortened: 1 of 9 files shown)


aria/cli/helptexts.py
```diff
@@ -47,3 +47,8 @@
 SORT_BY = "Key for sorting the list"
 DESCENDING = "Sort list in descending order [default: False]"
 JSON_OUTPUT = "Output logs in a consumable JSON format"
+
+DISPLAY_JSON = "Display in JSON format"
```
ran-z: lets merge this with `JSON_OUTPUT`,
the latter is a good name, and simply have it as "Format the output as JSON"

aria/cli/core/aria.py
```diff
@@ -325,6 +325,30 @@ def __init__(self):
             default=defaults.SERVICE_TEMPLATE_FILENAME,
             help=helptexts.SERVICE_TEMPLATE_FILENAME)
 
+        self.display_json = click.option(
```
ran-z: merge this with the other JSON one

aria/cli/core/aria.py
```diff
@@ -325,6 +325,30 @@ def __init__(self):
...
+            '-j',
+            '--json',
+            is_flag=True,
+            help=helptexts.DISPLAY_JSON)
+
+        self.display_yaml = click.option(
```
ran-z: `yaml_output`?

aria/cli/commands/services.py
```diff
@@ -177,3 +179,31 @@ def inputs(service_name, model_storage, logger):
...
+@aria.pass_logger
+def display(service_name, model_storage, json, yaml, graph, logger):
+    """Display information for a specific service template
+
+    `SERVICE_NAME` is the name of the service to display information on.
+    """
+    logger.info('Displaying service {0}...'.format(service_name))
+    service = model_storage.service.get_by_name(service_name)
+    consumption.ConsumptionContext()
+    if graph:
```
ran-z: General reminder that none of the methods below are standard for the CLI
Fine for now, but keep https://issues.apache.org/jira/browse/ARIA-183 in mind

aria/cli/commands/services.py
```diff
@@ -177,3 +179,31 @@ def inputs(service_name, model_storage, logger):
...
+@services.command(name='display',
+                  short_help='Display service information')
+@aria.argument('service-name')
+@aria.options.verbose()
+@aria.options.display_json
+@aria.options.display_yaml
+@aria.options.display_graph
+@aria.pass_model_storage
+@aria.pass_logger
+def display(service_name, model_storage, json, yaml, graph, logger):
```
ran-z: nitpicking that model_storage should probably sit after `graph`, before `logger`

aria/cli/commands/service_templates.py
```diff
@@ -181,21 +184,49 @@ def validate(service_template, service_template_filename,
...
+@service_templates.command(name='display',
+                           short_help='Display service template information')
+@aria.argument('service-template-name')
+@aria.options.verbose()
+@aria.options.display_json
+@aria.options.display_yaml
+@aria.options.display_types
+@aria.pass_model_storage
+@aria.pass_logger
+def display(service_template_name, model_storage, json, yaml, types, logger):
```
ran-z: Having both `display` and `show` is confusing :/
I agree however that it doesn't make sense to make this command into a mere flag of `show` as they're very different.
I'm generally ok with this as is, but Maxim wanted me to suggest `aria parse service/service-template` as an option as well. I'll leave it for your consideration :)

aria/cli/table.py
```diff
@@ -85,7 +85,7 @@ def get_values_per_column(column, row_data):
 
             return val
         else:
-            return defaults[column]
+            return defaults.get(column) if defaults is not None else None
```
ran-z: should instead have a `defaults = defaults or {}` before calling the nested method

aria/cli/commands/services.py
```diff
@@ -38,6 +41,53 @@ def services():
...
+def show(service_name, model_storage, all, graph, json, yaml, logger):
+    """Show information for a specific service template
+
+    `SERVICE_NAME` is the name of the service to display information on.
+    """
+    logger.info('Showing service {0}...'.format(service_name))
+    service = model_storage.service.get_by_name(service_name)
+
+    if json or yaml:
+        all = True
```
ran-z: i'd rename the flag, it's both unclear and a python keyword. `show_all` or `all_info` might be better.

ran-z: i actually didnt mean the flag but only the variables etc.,
although now that i think of it it might also be better to change the flag, as it might be confusing i.e. users might think it will show all service-templates or so..?
in any case what matters to me more is the variables, the flag can still be `all` to the user, but the "destination" of the variable (configurable in `aria.py`) should be renamed.

aria/cli/commands/service_templates.py
```diff
@@ -43,32 +45,51 @@ def service_templates():
...
-    if service_template_dict['description'] is not None:
-        logger.info('Description:')
-        logger.info('{0}{1}'.format(service_template_dict['description'].encode('UTF-8') or '',
-                                    os.linesep))
-
-    if service_template.services:
-        logger.info('Existing services:')
-        for service in service_template.services:
-            logger.info('\t{0}'.format(service.name))
+    if json or yaml:
```
ran-z: i'm a bit confused by this. so `yaml`, `json`, and `all` all mean all,
and `json` and `yaml` can be used simulatniously but only `json` will get printed.

Re the latter problem, please have a look at `aria.py` for how to make flags mutually-exclusive.
re the former, it might be better to also have the `all` flag as mutually-exclusive with the other two..

aria/cli/core/aria.py
```diff
@@ -325,6 +339,46 @@ def __init__(self):
             default=defaults.SERVICE_TEMPLATE_FILENAME,
             help=helptexts.SERVICE_TEMPLATE_FILENAME)
 
+        self.show_full = click.option(
+            '-f',
+            '--full',
+            'mode',
+            is_flag=True,
+            flag_value='full',
```
ran-z: I think I don't like this approach as much at the previous one.
It seems like it simplifies the code a bit but makes things somewhat more confusing to the user - specifically the fact that only the last flag of a given "destination" (e.g. `mode`) takes hold.

If instead you have different destinations for these flags, it'd be easier to inform the user of misuse (even if for some reason the mutually-exclusive mechanism is problematic to use, you can always just manually validate the parameters in the command function itself)

aria/cli/core/aria.py
```diff
@@ -40,16 +41,8 @@
...
-    with this class name and a `mutually_exclusive` argument with a list of
-    argument names it is mutually exclusive with.
-
-    NOTE: All mutually exclusive options must use this. It's not enough to
-    use it in just one of the options.
-    """
-
     def __init__(self, *args, **kwargs):
-        self.mutually_exclusive = set(kwargs.pop('mutually_exclusive', []))
+        self.mutually_exclusive = kwargs.pop('mutually_exclusive', tuple())
```
ran-z: why change this from a set to another type?

aria/cli/core/aria.py
```diff
@@ -65,15 +58,33 @@ def __init__(self, *args, **kwargs):
         super(MutuallyExclusiveOption, self).__init__(*args, **kwargs)
 
     def handle_parse_result(self, ctx, opts, args):
-        if self.mutually_exclusive.intersection(opts) and self.name in opts:
+        if (self.name in opts) and self.mutually_exclusive.keys().intersection(opts):
```
ran-z: seems like `mutually_exclusive` might be a `tuple` now, so `.keys()` would raise an error here.

aria/cli/core/aria.py
```diff
@@ -65,15 +58,33 @@ def __init__(self, *args, **kwargs):
...
                     self.mutuality_string,
                     self.mutuality_error_message))
-        return super(MutuallyExclusiveOption, self).handle_parse_result(
-            ctx, opts, args)
+        return super(MutuallyExclusiveOption, self).handle_parse_result(ctx, opts, args)
+
+
+def mutually_exclusive_option(*param_decls, **attrs):
+    """
+    Makes options mutually exclusive. The decorator must pass a a ``mutually_exclusive`` argument
```
ran-z: a a a a a

aria/cli/core/aria.py
```diff
@@ -65,15 +58,33 @@ def __init__(self, *args, **kwargs):
...
+    with a list of argument names with which the option is mutually exclusive.
+
+    NOTE: All mutually exclusive options must use this. It's not enough to use it in just one of the
+    options.
+    """
+    def decorator(func):
+        if 'help' in attrs:
```
ran-z: could you please document this func a bit more?
I'm not sure what are `param_decls` and `attrs` in this case, why `help` is being inspected, and why the internal `__click_params__` is accessed directly here.

PR #123, opened 2017-05-07
Dry execution changes the state of non implemented operations

mxmrlv:


aria/orchestrator/workflows/executor/base.py
```diff
@@ -31,7 +31,14 @@ def execute(self, task):
         Execute a task
         :param task: task to execute
         """
-        raise NotImplementedError
+        if task.model_task.implementation:
+            self._execute_empty_task(task)
+        else:
+            raise NotImplementedError
```
ran-z: ??
so to implement this i have to override `execute` entirely?

PR #125, opened 2017-05-07
ARIA-165 Make node name suffix UUIDs become more readable

mxmrlv:


aria/modeling/service_template.py
```diff
@@ -549,9 +549,13 @@ def as_raw(self):
...
 
     def instantiate(self, container):
-        context = ConsumptionContext.get_thread_local()
         from . import models
-        name = context.modeling.generate_node_id(self.name)
+        if self.nodes:
+            latest_node_index = self.nodes[-1].name.rsplit('_', 1)[1]
+            index = int(latest_node_index) + 1
+        else:
+            index = 0
```
ran-z: i'd start from index=1 actually

aria/modeling/service_template.py
```diff
@@ -549,9 +549,13 @@ def as_raw(self):
             ('requirement_templates', formatting.as_raw_list(self.requirement_templates))))
 
     def instantiate(self, container):
-        context = ConsumptionContext.get_thread_local()
         from . import models
-        name = context.modeling.generate_node_id(self.name)
+        if self.nodes:
+            latest_node_index = self.nodes[-1].name.rsplit('_', 1)[1]
```
ran-z: is the nodes list guaranteed to be ordered?
if not then we need to find the max

either way, i'd change `latest` to `highest`

mxmrlv: Pretty sure sql orders according to the indexed column

aria/modeling/service_template.py
```diff
@@ -549,9 +549,14 @@ def as_raw(self):
             ('requirement_templates', formatting.as_raw_list(self.requirement_templates))))
 
     def instantiate(self, container):
-        context = ConsumptionContext.get_thread_local()
         from . import models
-        name = context.modeling.generate_node_id(self.name)
+        import pydevd; pydevd.settrace('localhost', suspend=False)
```
ran-z: poo

aria/modeling/service_template.py
```diff
@@ -549,9 +549,14 @@ def as_raw(self):
...
+        if self.nodes:
+            highest_name_suffix = max(self.nodes, key=lambda n: n.name).name.rsplit('_', 1)[1]
```
ran-z: im not sure this would work;
note that `mynode_2` will come after `mynode_11` :|

PR #128, opened 2017-05-08
ARIA-210 Handle relative paths in CLI service-templates

AviaE:
This was a rather simple change, mainly involving adding absolute path
references.

The problems were only in `service-templates store` and in
`service-templates create-archive`.
`service-templates validate` was not affected.

### DIFF
--- a/aria/cli/commands/service_templates.py
+++ b/aria/cli/commands/service_templates.py
@@ -195,7 +195,9 @@ def create_archive(service_template_path, destination, logger):
     `destination` is the path of the output CSAR archive file
     """
     logger.info('Creating a CSAR archive')
-    csar.write(os.path.dirname(service_template_path), service_template_path, destination, logger)
+    if not destination.endswith(csar.CSAR_FILE_EXTENSION):
+        destination += csar.CSAR_FILE_EXTENSION
+    csar.write(service_template_path, destination, logger)
     logger.info('CSAR archive created at {0}'.format(destination))
 
 
--- a/aria/cli/csar.py
+++ b/aria/cli/csar.py
@@ -22,7 +22,7 @@ import zipfile
 import requests
 from ruamel import yaml
 
-
+CSAR_FILE_EXTENSION = '.csar'
 META_FILE = 'TOSCA-Metadata/TOSCA.meta'
 META_FILE_VERSION_KEY = 'TOSCA-Meta-File-Version'
 META_FILE_VERSION_VALUE = '1.0'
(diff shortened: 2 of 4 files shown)


aria/cli/csar.py
```diff
@@ -38,17 +38,19 @@
 }
 
 
-def write(source, entry, destination, logger):
-    source = os.path.expanduser(source)
-    destination = os.path.expanduser(destination)
-    entry_definitions = os.path.join(source, entry)
+def write(service_template_path, destination, logger):
```
ran-z: good, i like the signature better this way

aria/cli/commands/service_templates.py
```diff
@@ -195,7 +195,9 @@ def create_archive(service_template_path, destination, logger):
     `destination` is the path of the output CSAR archive file
     """
     logger.info('Creating a CSAR archive')
-    csar.write(os.path.dirname(service_template_path), service_template_path, destination, logger)
+    if not destination.endswith('.csar'):
```
ran-z: maybe extract `.csar` suffix to a constant in csar.py? seems like its defined in several places now

aria/cli/service_template_utils.py
```diff
@@ -53,7 +53,7 @@ def get(source, service_template_filename):
             return _get_service_template_file_from_archive(source, service_template_filename)
```
ran-z: What about adding a short test or two for these scenarios? :)

tests/cli/test_service_templates.py
```diff
@@ -244,3 +258,13 @@ def test_create_archive_successful(self, monkeypatch, mock_object):
         monkeypatch.setattr(csar, 'write', mock_object)
         self.invoke('service_templates create_archive stubpath stubdest')
         assert 'CSAR archive created at stubdest' in self.logger_output_string
+
+    def test_create_archive_from_relative_path(self, monkeypatch, mock_object):
+
+        monkeypatch.setattr(os.path, 'isfile', lambda x: True)
+        monkeypatch.setattr(os.path, 'isfile', mock_object)
```
ran-z: doesnt this line override the previous one?

aria/cli/csar.py
```diff
@@ -22,7 +22,7 @@
 import requests
 from ruamel import yaml
 
-
+CSAR_FILE_EXTENSION = '.csar'
```
ran-z: do a quick search for `.csar`, there should be at least one more reference to it in modules you've touched anyway :)

tests/cli/test_service_templates.py
```diff
@@ -131,6 +133,18 @@ def test_store_no_exception(self, monkeypatch, mock_object):
...
 
+    def test_store_relative_path_single_yaml_file(self, monkeypatch, mock_object):
+        monkeypatch.setattr(Core, 'create_service_template', mock_object)
+        monkeypatch.setattr(os.path, 'isfile', lambda x: True)
+        monkeypatch.setattr(service_template_utils, '_is_archive', lambda x: False)
+
+        self.invoke('service_templates store service_template.yaml {name}'.format(
+            name=mock_models.SERVICE_TEMPLATE_NAME))
+
+        mock_object.assert_called_with(os.path.join(os.getcwd(), 'service_template.yaml'),
```
ran-z: just curious why you've decided to use `join` and `getcwd` here rather than absolute path (like you did in the actual code). not that it should matter :)

aria/cli/service_template_utils.py
AviaE: Added tests.

aria/cli/csar.py
AviaE: I did before, these were the only places =)

tests/cli/test_service_templates.py
AviaE: Nothing serious, just thought that it demonstrates the fact that we are expecting the paths to be relative to the current working directory.

tests/cli/test_service_templates.py
AviaE: Indeed. So I removed that line.

PR #138, opened 2017-05-22
ARIA-149 Enhance operation configuration

tliron:
* Parse special "dependencies" configuration parameters as YAML and
  treat as Parameter models, allowing them full use of intrinsic
  functions, type coersions, and validations
* Rename various functions that process "properties" to more generically
  process "parameters" (properties, inputs, attributes, arguments, etc.)
* The "configuration" field in OperationTemplate and Operation models
  is now now a dict of Parameter models
* Added "function" and "arguments" fields to Operation model to preserve
  user data (in "implementation" and "inputs") and to clearly demarcate
  orchestration data from user data
* Some cleanup of parser code touched by this commit


aria/modeling/exceptions.py
```diff
@@ -22,9 +22,9 @@ class ModelingException(AriaException):
     """
 
 
-class InputsException(ModelingException):
+class ParameterException(ModelingException):
```
ran-z: I'm not sure I understand this change - It's true that the model object is `Parameter` and not `Inputs` (although that too is supposedly under change by @AviaE ) - but the error itself is not about the model type and doesn't have anything much to do with the models themselves - it's about inputs, so why shouldn't the exception be named like so?

tests/resources/service-templates/tosca-simple-1.0/node-cellar/node-cellar.yaml
```diff
@@ -309,7 +311,7 @@ policy_types:
       client connections cleanly and shut down services. 
     derived_from: aria.Workflow
     properties:
-      implementation:
+      function:
```
ran-z: i thought in the service-template even according to your view this should remain `implementation`, no?

extensions/aria_extension_tosca/simple_v1_0/modeling/parameters.py
```diff
@@ -71,6 +72,7 @@ def get_assigned_and_defined_property_values(context, presentation, field_name='
     values = OrderedDict()
 
     the_type = presentation._get_type(context)
+    field_name_plural = pluralize(field_name)
```
ran-z: how did the following lines worked before this? o_O

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
```diff
@@ -384,18 +387,37 @@ def create_operation_template_model(context, service_template, operation):
...
+                    # Parse as YAML
+                    try:
+                        value = yaml.load(value)
+                    except yaml.parser.MarkedYAMLError as e:
+                        context.validation.report(
+                            'YAML parser {0} in operation configuration: {1}'
+                            .format(e.problem, value),
+                            locator=implementation._locator,
+                            level=Issue.FIELD)
+                        continue
```
ran-z: shouldn't this be a re-raise? seems somewhat fatal...

aria/orchestrator/execution_plugin/instantiation.py
```diff
@@ -16,107 +16,132 @@
...
-    ssh = _get_ssh(configuration.pop('ssh')) if 'ssh' in configuration else {}
     if 'user' not in ssh:
         ssh['user'] = default_user
     if ('password' not in ssh) and ('key' not in ssh) and ('key_filename' not in ssh):
         ssh['password'] = default_password
 
-    arguments['use_sudo'] = ssh.get('use_sudo', False)
-    arguments['hide_output'] = ssh.get('hide_output', [])
-    arguments['fabric_env'] = {}
+    operation.arguments['use_sudo'] = Parameter.wrap('use_sudo', ssh.get('use_sudo', False),
```
ran-z: shouldn't the wrapping of configuration parameters simply take place where this module is called from?
what about configuration parameters not meant for the execution plugin?

aria/modeling/service_instance.py
```diff
@@ -1728,34 +1732,41 @@ def operation_template(cls):
...
-            # can validate the configuration and otherwise return specially derived arguments
-            arguments = self.configuration
+            # can validate the configuration and otherwise create specially derived arguments. For
+            # now, we just send all configuration parameters as arguments
+            utils.instantiate_dict(self, self.arguments, self.configuration)
 
-        # Note: the arguments will *override* operation inputs of the same name
-        if arguments:
-            for k, v in arguments.iteritems():
-                self.inputs[k] = models.Parameter.wrap(k, v)
```
ran-z: why remove this?

ran-z: :+1:

aria/modeling/exceptions.py
tliron: Discussed and we agreed that more general is better in this case.

tests/resources/service-templates/tosca-simple-1.0/node-cellar/node-cellar.yaml
tliron: Correct, changed back.

extensions/aria_extension_tosca/simple_v1_0/modeling/parameters.py
tliron: Previously we sent the plural as a function argument, but I optimized by removing it... just some code cleanup.

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
tliron: Validation issues accumulate and are displayed in the end. None are considered fatal enough to start the parsing, but of course a single error means that parsing failed.

aria/orchestrator/execution_plugin/instantiation.py
tliron: We went through the code together, this happens elsewhere.

aria/modeling/utils.py
```diff
@@ -52,84 +53,95 @@ def service_template(self):
         return self.container.service_template
 
 
-def create_parameters(parameters, declared_parameters):
+def merge_parameter_values(parameter_values, declared_parameters, forbidden_names=None):
```
ran-z: i thought we talked about the concept of validating "forbidden names" and we agreed it should be done in the parser / instantiation phases instead. I don't see in what scenario we'd like to inform the user of such validation errors only at this stage.

aria/modeling/exceptions.py
```diff
@@ -57,3 +57,9 @@ class UndeclaredParametersException(ParameterException):
     """
     ARIA modeling exception: Undeclared parameters have been provided.
     """
+
+
+class ForbiddenParameterNamesException(ParameterException):
```
ran-z: perhaps "Reserved" is more appropriate than "Forbidden"?

aria/orchestrator/execution_plugin/instantiation.py
ran-z: It's true that configuration parameters are wrapped in another place when it's not intended for the execution plugin, but my question is why bother wrapping them at this stage at all - as opposed to working with dicts at this stage and then have the instantiation code which calls this module do the wrapping. it just makes more sense IMO.

aria/modeling/exceptions.py
ran-z: :+1:

extensions/aria_extension_tosca/simple_v1_0/modeling/parameters.py
ran-z: :+1:

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
ran-z: :+1:

tests/resources/service-templates/tosca-simple-1.0/node-cellar/node-cellar.yaml
ran-z: :+1:

tests/orchestrator/test_workflow_runner.py
```diff
@@ -259,8 +258,9 @@ def _setup_mock_workflow_in_service(request, inputs=None):
     workflow = models.Operation(
         name=mock_workflow_name,
         service=service,
-        implementation='workflow.mock_workflow',
-        inputs=inputs or {})
+        function='workflow.mock_workflow',
+        inputs=inputs or {},
+        arguments=inputs or {})
```
ran-z: why is this needed? wouldn't arguments get auto-populated with inputs anyway?

aria/orchestrator/execution_plugin/instantiation.py
tliron: If you notice I am also giving the parameters descriptions here. These are not mere dicts, but true parameters.

aria/orchestrator/workflow_runner.py
```diff
@@ -136,10 +137,11 @@ def _validate_no_active_executions(self, execution):
         active_executions = [e for e in self.service.executions if e.is_active()]
         if active_executions:
             raise exceptions.ActiveExecutionsError(
-                "Can't start execution; Service {0} has an active execution with id {1}"
+                "Can't start execution; Service {0} has an active execution with ID {1}"
                 .format(self.service.name, active_executions[0].id))
 
-    def _get_workflow_fn(self):
+    @property
+    def _workflow_fn(self):
```
tliron: Changing back to function.

aria/modeling/utils.py
```diff
@@ -51,74 +53,95 @@ def service_template(self):
         return self.container.service_template
 
 
-def create_inputs(inputs, template_inputs):
+def merge_parameter_values(parameter_values, declared_parameters, forbidden_names=None):
```
ran-z: what was the purpose of the changes to this method? im not sure i understand the benefit and it seems less readable now to me

aria/orchestrator/workflows/events_logging.py
```diff
@@ -35,20 +35,20 @@ def _get_task_name(task):
...
 def _start_task_handler(task, **kwargs):
-    # If the task has not implementation this is an empty task.
-    if task.implementation:
+    # If the task has no function this is an empty task.
+    if task.function:
         suffix = 'started...'
         logger = task.context.logger.info
     else:
-        suffix = 'has no implementation'
+        suffix = 'has no function'
```
tliron: Changing back to "implementation".

extensions/aria_extension_tosca/simple_v1_0/modeling/functions.py
```diff
@@ -69,7 +69,7 @@ def __evaluate__(self, container_holder):
             e, final = evaluate(e, final, container_holder)
             if e is not None:
                 value.write(unicode(e))
-        value = value.getvalue()
+        value = value.getvalue() or u''
```
tliron: We default to Unicode everywhere possible in values -- this avoids thorny problems later down the road with values.

aria/modeling/exceptions.py
tliron: Changing to "Reserved".

tests/orchestrator/test_workflow_runner.py
tliron: No, not in the way this particular test works, since it creates an operation model directly and doesn't call `configure()`

aria/modeling/utils.py
tliron: The original code was superfluous and complex, doing unpacking and re-packing of values, and actually had a bug in one area of it. I think this code is much cleaner!

tests/orchestrator/test_workflow_runner.py
ran-z: right :+1:

aria/modeling/exceptions.py
ran-z: :+1:

extensions/aria_extension_tosca/simple_v1_0/modeling/functions.py
ran-z: :+1:

aria/orchestrator/workflows/events_logging.py
ran-z: :+1:

aria/orchestrator/workflow_runner.py
ran-z: :+1:

aria/modeling/utils.py
ran-z: :+1:

aria/modeling/utils.py
ran-z: :+1:

PR #139, opened 2017-05-25
ARIA-261 Single-source ARIA version

ran-z:


setup.py
```diff
@@ -119,6 +120,9 @@ class DevelopCommand(develop):
         'License :: OSI Approved :: Apache Software License',
         'Operating System :: OS Independent',
         'Programming Language :: Python',
+        'Programming Language :: Python :: 2',
```
johnament: If you can please fix the license statement to match thrift's example https://github.com/apache/thrift/blob/master/lib/py/setup.py#L95

ran-z: :+1:

PR #143, opened 2017-06-01
ARIA-254 Create multiple nodes per template

tliron:
* New aria.Scaling policy (and "scaling" role)
* NodeTemplate model no longer stores scaling values (default_instances,
etc.) but instead fetches them from applicable scaling policies
* Some code cleanup

### DIFF
--- a/aria/modeling/service_template.py
+++ b/aria/modeling/service_template.py
@@ -65,34 +65,6 @@ class ServiceTemplateBase(TemplateModelMixin):
                           'interface_type_fk',
                           'artifact_type_fk')
 
-    description = Column(Text, doc="""
-    Human-readable description.
-
-    :type: :obj:`basestring`
-    """)
-
-    main_file_name = Column(Text, doc="""
-    Filename of CSAR or YAML file from which this service template was parsed.
-    
-    :type: :obj:`basestring`
-    """)
-
-    created_at = Column(DateTime, nullable=False, index=True, doc="""
-    Creation timestamp.
-
-    :type: :class:`~datetime.datetime`
-    """)
-
-    updated_at = Column(DateTime, doc="""
-    Update timestamp.
-
-    :type: :class:`~datetime.datetime`
-    """)
-
-    # region association proxies
-
-    # endregion
-
     # region one_to_one relationships
 
     @declared_attr
@@ -253,10 +225,6 @@ class ServiceTemplateBase(TemplateModelMixin):
 
     # endregion
 
-    # region many_to_one relationships
-
-    # endregion
-
     # region many_to_many relationships
 
     @declared_attr
@@ -315,6 +283,30 @@ class ServiceTemplateBase(TemplateModelMixin):
 
     # endregion
 
+    description = Column(Text, doc="""
+    Human-readable description.
+
+    :type: :obj:`basestring`
+    """)
+
+    main_file_name = Column(Text, doc="""
+    Filename of CSAR or YAML file from which this service template was parsed.
+    
+    :type: :obj:`basestring`
+    """)
+
+    created_at = Column(DateTime, nullable=False, index=True, doc="""
+    Creation timestamp.
+
+    :type: :class:`~datetime.datetime`
+    """)
+
+    updated_at = Column(DateTime, doc="""
+    Update timestamp.
+
+    :type: :class:`~datetime.datetime`
+    """)
+
     @property
     def as_raw(self):
         return collections.OrderedDict((
@@ -341,12 +333,14 @@ class ServiceTemplateBase(TemplateModelMixin):
 
     def instantiate(self, container, model_storage, inputs=None):  # pylint: disable=arguments-differ
         from . import models
-        context = ConsumptionContext.get_thread_local()
         now = datetime.now()
         service = models.Service(created_at=now,
                                  updated_at=now,
                                  description=deepcopy_with_locators(self.description),
                                  service_template=self)
+
+        # TODO: we want to remove this use of the context
+        context = ConsumptionContext.get_thread_local()
         context.modeling.instance = service
 
         service.inputs = utils.merge_parameter_values(inputs, self.inputs, model_cls=models.Input)
@@ -365,7 +359,7 @@ class ServiceTemplateBase(TemplateModelMixin):
         utils.instantiate_dict(self, service.meta_data, self.meta_data)
 
         for node_template in self.node_templates.itervalues():
-            for _ in range(node_template.default_instances):
+            for _ in range(node_template.scaling['default_instances']):
                 node = node_template.instantiate(container)
                 service.nodes[node.name] = node
 
@@ -468,22 +462,6 @@ class NodeTemplateBase(TemplateModelMixin):
     __private_fields__ = ('type_fk',
                           'service_template_fk')
 
-    # region association proxies
-
-    @declared_attr
-    def service_template_name(cls):
-        return relationship.association_proxy('service_template', 'name')
-
-    @declared_attr
-    def type_name(cls):
-        return relationship.association_proxy('type', 'name')
-
-    # endregion
-
-    # region one_to_one relationships
-
-    # endregion
-
     # region one_to_many relationships
 
     @declared_attr
@@ -573,6 +551,18 @@ class NodeTemplateBase(TemplateModelMixin):
 
     # endregion
 
+    # region association proxies
+
+    @declared_attr
+    def service_template_name(cls):
+        return relationship.association_proxy('service_template', 'name')
+
+    @declared_attr
+    def type_name(cls):
+        return relationship.association_proxy('type', 'name')
+
+    # endregion
+
     # region foreign_keys
 
     @declared_attr
@@ -617,22 +607,12 @@ class NodeTemplateBase(TemplateModelMixin):
     :type: [:class:`NodeTemplateConstraint`]
     """)
 
-    def is_target_node_template_valid(self, target_node_template):
-        if self.target_node_template_constraints:
-            for node_template_constraint in self.target_node_template_constraints:
-                if not node_template_constraint.matches(self, target_node_template):
-                    return False
-        return True
-
     @property
     def as_raw(self):
         return collections.OrderedDict((
             ('name', self.name),
             ('description', self.description),
             ('type_name', self.type.name),
-            ('default_instances', self.default_instances),
-            ('min_instances', self.min_instances),
-            ('max_instances', self.max_instances),
             ('properties', formatting.as_raw_dict(self.properties)),
             ('attributes', formatting.as_raw_dict(self.properties)),
             ('interface_templates', formatting.as_raw_list(self.interface_templates)),
@@ -642,13 +622,7 @@ class NodeTemplateBase(TemplateModelMixin):
 
     def instantiate(self, container):
         from . import models
-        if self.nodes:
-            highest_name_suffix = max(int(n.name.rsplit('_', 1)[-1]) for n in self.nodes)
-            suffix = highest_name_suffix + 1
-        else:
-            suffix = 1
-        name = '{name}_{index}'.format(name=self.name, index=suffix)
-        node = models.Node(name=name,
+        node = models.Node(name=self._next_name,
                            type=self.type,
                            description=deepcopy_with_locators(self.description),
                            state=models.Node.INITIAL,
@@ -660,10 +634,12 @@ class NodeTemplateBase(TemplateModelMixin):
         utils.instantiate_dict(node, node.capabilities, self.capability_templates)
 
         # Default attributes
-        if 'tosca_name' in node.attributes:
+        if ('tosca_name' in node.attributes) \
+            and (node.attributes['tosca_name'].type_name == 'string'):
             node.attributes['tosca_name'].value = self.name
-        if 'tosca_id' in node.attributes:
-            node.attributes['tosca_id'].value = name
+        if 'tosca_id' in node.attributes \
+            and (node.attributes['tosca_id'].type_name == 'string'):
+            node.attributes['tosca_id'].value = node.name
 
         return node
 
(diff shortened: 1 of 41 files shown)


tests/end2end/test_nodecellar.py
```diff
@@ -39,4 +39,4 @@ def _verify_deployed_service_in_storage(service_name, model_storage):
     service = service_templates[0].services[service_name]
     assert service.name == service_name
     assert len(service.executions) == 0  # dry executions leave no traces
-    assert len(service.nodes) == 10
+    assert len(service.nodes) == 13
```
mxmrlv: why 13? I though you added just one other instance

extensions/aria_extension_tosca/simple_v1_0/types.py
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from aria.utils.collections import FrozenDict, FrozenList
+from aria.utils.collections import (FrozenDict, FrozenList)
```
mxmrlv: if adding parentheses, why not try and follow the style guide? i.e. `from aria.utils import collections`

aria/modeling/service_template.py
```diff
@@ -613,6 +597,56 @@ def dump(self):
...
+    @property
+    def default_instances(self):
+        # TODO: currently finds the first matching policy; but we should emit a validation error
+        # if more than one policy applies to the same node
+        for policy_template in self.service_template.policy_templates.itervalues():
+            if policy_template.type.role == 'scaling':
+                if policy_template.is_for_node_template(self.name):
+                    default_instances = policy_template.properties.get('default_instances')
+                    if (default_instances is not None) \
+                        and (default_instances.type_name == 'integer'):
```
mxmrlv: how can it not be 'integer'?

tests/end2end/test_nodecellar.py
tliron: The policy is on the `node_cellar_group` which contains three 3 templates. So we get 3 extra nodes.

extensions/aria_extension_tosca/simple_v1_0/types.py
tliron: I thought we agreed that that kind of style is a recommendation but not a requirement. If we switch to that style here it would make a lot of the parser code extremely verbose and frankly unreadable.

aria/modeling/service_template.py
tliron: We cannot expect anybody be to using the ARIA profile as is -- all we are relying here is that there is an attribute with that name. I am adding that extra validation of type to make sure that we don't cause breakage by somebody messing with the parameters. Note that we also do this validation in setting "default" attributes, like `tosca_name` and `tosca_id`.

extensions/aria_extension_tosca/simple_v1_0/types.py
mxmrlv: if by adding the module prefix makes the code unreadable, maybe we need to refactor the structure?

tests/end2end/test_nodecellar.py
mxmrlv: kewl

extensions/aria_extension_tosca/profiles/tosca-simple-1.0/capabilities.yaml
```diff
@@ -132,6 +132,7 @@ capability_types:
       specification: tosca-simple-1.0
       specification_section: 5.4.10
       specification_url: 'http://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.0/cos01/TOSCA-Simple-Profile-YAML-v1.0-cos01.html#DEFN_TYPE_CAPABILITIES_SCALABLE'
+      role: scaling
```
mxmrlv: what does it mean to have a scaling role?

extensions/aria_extension_tosca/simple_v1_0/modeling/capabilities.py
```diff
@@ -162,6 +164,30 @@ def convert_capability_from_definition_to_assignment(context, presentation, cont
...
+    merge_raw_parameter_definitions(context, presentation, raw_properties,
+                                    capability_definition.properties, 'properties')
+
+    if raw_properties:
+        capability_definition._raw['properties'] = raw_properties
+        capability_definition._reset_method_cache()
+
+    # Merge occurrences
+    occurrences = from_capability_definition._raw.get('occurrences')
+    if (occurrences is not None) and (capability_definition._raw.get('occurrences') is None):
```
mxmrlv: why the redundant parenthesis?

extensions/aria_extension_tosca/simple_v1_0/modeling/capabilities.py
```diff
@@ -52,12 +52,13 @@ def get_inherited_capability_definitions(context, presentation, for_presentation
...
     parent = presentation._get_parent(context)
-    capability_definitions = get_inherited_capability_definitions(context, parent,
-                                                                  for_presentation=presentation) \
-                                                                  if parent is not None \
-                                                                  else OrderedDict()
+    capability_definitions = get_inherited_capability_definitions(
+        context, parent, for_presentation) if parent is not None else OrderedDict()
 
     # Add/merge our capability definitions
     our_capability_definitions = presentation.capabilities
```
mxmrlv: it's really hard to follow with name like 
capability_definitions and our_capability_definitions

extensions/aria_extension_tosca/simple_v1_0/modeling/capabilities.py
```diff
@@ -162,6 +164,30 @@ def convert_capability_from_definition_to_assignment(context, presentation, cont
     return CapabilityAssignment(name=presentation._name, raw=raw, container=container)
 
 
+def merge_capability_definition(context, presentation, capability_definition,
+                                from_capability_definition):
```
mxmrlv: the hard to follow comment applies here as well

extensions/aria_extension_tosca/simple_v1_0/modeling/capabilities.py
```diff
@@ -162,6 +164,30 @@ def convert_capability_from_definition_to_assignment(context, presentation, cont
...
+    raw_properties = OrderedDict()
+
+    # Merge properties from type
+    from_property_defintions = from_capability_definition.properties
+    merge_raw_parameter_definitions(context, presentation, raw_properties, from_property_defintions,
+                                    'properties')
+
+    # Merge our properties
+    merge_raw_parameter_definitions(context, presentation, raw_properties,
+                                    capability_definition.properties, 'properties')
```
mxmrlv: if i get this right, capability_definitions is the "global" definition of this capability, while "from..." if the definitions for this instance. 
and we first apply our own definitions and only then try and merge the global one - while maintaining the idea that the more specific definitions will be the ones to hold?

aria/modeling/service_template.py
```diff
@@ -690,19 +666,104 @@ def dump(self):
...
-                else ' or more'))
             utils.dump_dict_values(self.properties, 'Properties')
             utils.dump_dict_values(self.attributes, 'Attributes')
             utils.dump_interfaces(self.interface_templates)
             utils.dump_dict_values(self.artifact_templates, 'Artifact templates')
             utils.dump_dict_values(self.capability_templates, 'Capability templates')
             utils.dump_list_values(self.requirement_templates, 'Requirement templates')
 
+    @property
+    def next_index(self):
```
mxmrlv: cannot remain public

aria/modeling/service_template.py
```diff
@@ -690,19 +666,104 @@ def dump(self):
...
+        :rtype: int
+        """
+
+        max_index = 0
+        if self.nodes:
+            max_index = max(int(n.name.rsplit('_', 1)[-1]) for n in self.nodes)
+        return max_index + 1
+
+    @property
+    def next_name(self):
```
mxmrlv: so is this

mxmrlv: All the new functions are seem to be for orchestrator use only. Why would the user need to use these?

aria/modeling/service_template.py
```diff
@@ -690,19 +666,104 @@ def dump(self):
...
+                    extract_properties(policy_template.properties)
+
+        # Defaults
+        default_property('min_instances', 0)
+        default_property('max_instances', 1)
+        default_property('default_instances', 1)
+
+        # Validate
+        # pylint: disable=too-many-boolean-expressions
+        if (scaling['min_instances'] < 0) or \
```
mxmrlv: may any? and remove the pylint issue?

aria/modeling/service_template.py
```diff
@@ -690,19 +666,104 @@ def dump(self):
...
+            if name not in scaling:
+                scaling[name] = value
+
+        # From our scaling capabilities
+        for capability_template in self.capability_templates.itervalues():
+            if capability_template.type.role == 'scaling':
+                extract_properties(capability_template.properties)
+
+        # From service scaling policies
+        for policy_template in self.service_template.policy_templates.itervalues():
```
mxmrlv: so are we overriding the more specific properties with the policy based properties?

extensions/aria_extension_tosca/profiles/tosca-simple-1.0/capabilities.yaml
tliron: The `role` meta data is the way to specify special behavior for types in ARIA. This is much safer than checking if name == `aria.Scaling`, or even if it inherits from that. Roles are inherited with the type, so subtypes will have the same role (unless they override it). We use this in a few cases, for example there is `host` role for Compute and for HostedOn relationships.

extensions/aria_extension_tosca/simple_v1_0/modeling/capabilities.py
tliron: I tend to agree. :( This is some very tricky code that has gone through a lot of changes. I do want to fix this (there are a bunch of related files in this module) but I don't think this PR is the right place.

extensions/aria_extension_tosca/simple_v1_0/modeling/capabilities.py
tliron: I do not consider them redundant at all. Nobody should have to remember operation precedence, you should be able to tell at a glance. I recommend everybody follow this practice in every language. :)

extensions/aria_extension_tosca/simple_v1_0/modeling/capabilities.py
tliron: +1, but I suggest refactoring some other time

aria/modeling/service_template.py
tliron: Actually, this was intentional -- I thought it could be useful for outside callers. But no problem changed this and also `_next_name` similarly.

extensions/aria_extension_tosca/simple_v1_0/modeling/capabilities.py
tliron: Not exactly. This is a utility method called by `get_inherited_capability_definitions`. Its goal it to merge the current node type's capability definitions over the inherited parent node type's capability definitions. "from" here is the parent's.

aria/modeling/service_template.py
tliron: No, the opposite. There is no overriding: the `extract_property` function will not extract if the value is already there. First to come wins. :)

aria/modeling/service_template.py
tliron: I'm not sure how `any` would work here. How would you write this?

extensions/aria_extension_tosca/simple_v1_0/modeling/capabilities.py
mxmrlv: well i do agree when it comes to an elaborate conditions :)

extensions/aria_extension_tosca/profiles/tosca-simple-1.0/capabilities.yaml
mxmrlv: :+1:

extensions/aria_extension_tosca/simple_v1_0/modeling/capabilities.py
mxmrlv: I do get the logic of the entire method, i was referring to the order of the `merge_raw_parameter_definitions`...

aria/modeling/service_template.py
mxmrlv: Oh, didn't notice it. I would think that there is not real need to check, just override the inherited properties. But this is my personal preference

aria/modeling/service_template.py
mxmrlv:
```
if any([
 scaling['min_instances'] < 0,
 scaling['max_instances'] < 0,
 scaling['default_instances'] < 0,
 scaling['max_instances'] < scaling['min_instances'],
 scaling['default_instances'] < scaling['min_instances'],
 scaling['default_instances'] > scaling['max_instances']
]):
...
```

extensions/aria_extension_tosca/simple_v1_0/modeling/capabilities.py
mxmrlv: well, i might go with it regarding the first naming issue, but this is a new function, why not rename its arguments properly?

aria/modeling/service_template.py
tliron: I see no difference in readability between this and using `or`, seems to me just to be bowing to PyLint's weirdness... but I don't mind switching.

extensions/aria_extension_tosca/simple_v1_0/modeling/capabilities.py
tliron: Because I prefer it to fit with the style of the rest of code right now, to keep it consistent. I would like to change it, but everything together and give it some more thought.

PR #145, opened 2017-06-02
ARIA-260 Send interface inputs as arguments

tliron:


tests/end2end/test_nodecellar.py
```diff
@@ -40,3 +40,21 @@ def _verify_deployed_service_in_storage(service_name, model_storage):
     assert service.name == service_name
     assert len(service.executions) == 0  # dry executions leave no traces
     assert len(service.nodes) == 10
+    assert set(service.nodes['loadbalancer_host_1'].interfaces['Standard'].operations['create'] \
```
ran-z: While these additions are fine, I think this JIRA really calls for unit tests which will check the various options for passing in arguments and verify everything overrides what it should and only that.
There are many such scenarios, and you don't want to test them all in e2e tests.

tliron: I created a new tests/instantiation package and put them there for now.

tests/instantiation/__init__.py
```diff
@@ -0,0 +1,14 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
ran-z: +1 for `instantiation` tests package -  even though we dont have an `instanitation` code package yet, this is where I too think the tests should reside

tests/instantiation/test_configuration.py
```diff
@@ -0,0 +1,48 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+
+from tests.parser.service_templates import consume_node_cellar
```
ran-z: You should use a custom example here rather than the nodecellar one IMO.
For once, these tests shouldn't break when the nodecellar example gets updated for some reason.
Another reason is that it's simply not comprehensive enough - there are two unit tests in this module, but a lot more scenarios to cover, for example:
  1) interface input which gets overridden by an operation input
  2) interface input which gets overriden by both an operation input and a configuration input
  3) interface input with a default value in the `interface_types` section without an override in the node template's interface section for that input
...

Even if we don't cover all of them, we should cover most, and this is not something you want to have as part of a single main example, but rather lots of small yaml sections each intended for a single purpose test.

tests/resources/service-templates/tosca-simple-1.0/node-cellar/node-cellar.yaml
```diff
@@ -102,8 +114,10 @@ topology_template:
               #token: { token: [ 'zero.one|two-three', '.|-', 3 ] }
       interfaces:
         Maintenance:
-          enable: juju > charm.maintenance_on
-          disable: juju > charm.maintenance_off
+          inputs:
+            mode: eventual
+          enable: maintenance_node_cellar.sh
```
ran-z: why make this change?

tests/instantiation/test_configuration.py
```diff
@@ -0,0 +1,175 @@
...
+    yield context.modeling.instance
+
+
+@pytest.fixture
+def broken_service_issues():
+    context, _ = consume_literal(BROKEN_TEMPLATE, no_issues=False)
+    yield context.validation.issues
+
+
+def _values(the_dict):
```
ran-z: we have this line in so many places nowadays, maybe you can find some common place to extract it to?

tests/instantiation/test_configuration.py
```diff
@@ -0,0 +1,175 @@
...
+
+import pytest
+
+from tests.parser.service_templates import consume_literal
+
+
+TEMPLATE = """
+tosca_definitions_version: tosca_simple_yaml_1_0
+
+interface_types:
```
ran-z: This seems like it covers the basic case, but there are many other scenarios which we might want to test as well, e.g.:
1) interface type input overridden by an operation input defined under a node type and not in the interface
2) interface type input overridden by an operation input in an interface type inheriting from this interface type
3) interface type input overridden by an operation input in an interface type overridden by an operation input under a node type overridden by an operation input under a node template
...

tests/resources/service-templates/tosca-simple-1.0/node-cellar/node-cellar.yaml
tliron: I always want to keep the Node Cellar test resource as "rich" as possible in all features we support, as a sanity check. If we break something, Node Cellar will almost always fail.

tests/instantiation/test_configuration.py
tliron: On second look, I disagree with your suggested test cases. The `instantiation` testing is supposed to test instantiation, and what you're suggesting is parsing. Especially when we introduce a real instantiation module soon, I think such tests would look very strange here.

We obviously need to test the things you mention, but they are part of parser testing, a huge undertaking which we have yet to begin...

tests/instantiation/test_configuration.py
tliron: Done. `aria.modeling.utils.parameters_as_values`

PR #146, opened 2017-06-02
ARIA-199 Add "services outputs" CLI command

tliron:


tests/resources/service-templates/tosca-simple-1.0/node-cellar/node-cellar.yaml
```diff
@@ -302,6 +306,14 @@ topology_template:
     capabilities:
       app_endpoint: [ loadbalancer, client ]
 
+  outputs:
```
ran-z: It'd be better (or rather, also good) to have an `outputs` section appended to the `hello-world` example, and then use it in the end2end test to grab the URL of the web server.

tests/mock/models.py
```diff
@@ -96,6 +97,9 @@ def create_service_with_dependencies(include_execution=False,
     if include_input:
         input = create_parameter(name='input1', value='value1')
         service.inputs = {'input1': input}
+    if include_output:
```
AviaE: Now, that ARIA-180 was merged, we have a dedicated `create_input` function.
Similarly, please add a `create_output` function =)
In addition, as in `services outputs` we also print the description of the output, so it would be nice to add the description as a parameter to the `create_output` function, and to check for the existence of that printed description in the console output (`logger output string`).

On another note, According to the TOSCA spec, inputs also have a description, and the is not addressed in the current `create_input` and tests for `services outputs`. You have me permission [:)] to fix it in the context of this commit, as it is very closely related to it. If you prefer not to, it will be addressed in this JIRA: https://issues.apache.org/jira/browse/ARIA-192 ("Fix output inconsistencies in CLI").

tests/cli/test_services.py
```diff
@@ -174,7 +174,30 @@ def test_delete_available_nodes_error_with_force(self, monkeypatch, mock_storage
...
+    def test_outputs_one_output(self, monkeypatch, mock_storage):
+        monkeypatch.setattr(_Environment, 'model_storage', mock_storage)
+        s = mock_models.create_service_with_dependencies(include_output=True)
+        monkeypatch.setattr(mock_storage.service, 'get_by_name', mock.MagicMock(return_value=s))
+
+        self.invoke('services outputs test_s')
+
+        assert 'output1' in self.logger_output_string
+        assert 'value1' in self.logger_output_string
+        assert 'No inputs' not in self.logger_output_string
```
AviaE: `'No outputs'`

aria/cli/commands/services.py
```diff
@@ -192,17 +192,17 @@ def outputs(service_name, model_storage, logger):
...
-        outputs_.write('     Description: {0}{1}'.format(description,
-                                                         os.linesep))
-        outputs_.write('     Value: {0}{1}'.format(output, os.linesep))
-    logger.info(outputs_.getvalue())
+
+    if service.outputs:
+        outputs_ = StringIO()
+        for output_name, output in service.outputs.iteritems():
+            outputs_.write(' - "{0}":{1}'.format(output_name, os.linesep))
+            outputs_.write('     Description: {0}{1}'.format(output.description,
```
AviaE: You can put `os.linesep` in the same line as `Description`, it is short enough, and consistent with the other lines.

tests/mock/models.py
tliron: I think this should be done separately -- I'm trying to make this commit specifically about the CLI.

tests/resources/service-templates/tosca-simple-1.0/node-cellar/node-cellar.yaml
tliron: Done.

PR #157, opened 2017-06-20
ARIA-281 Update click library version

ran-z:
Updated click library version to a newer version.
This will also make it compatible in the same environment with the
pip-tools library which uses click too and requires the newer version.

Additional changes:
 - Removed PyYAML dependency from requirements.txt (leftover)
 - Disabled default goal of Makefile


asfgit:

PR #160, opened 2017-06-21
ARIA-282 Make SSH capability opt-in

ran-z:
Since the Fabric library uses Paramiko, which is a library
using a license which is incompatible with Apache's,
ARIA's SSH capabilities are now opt-in and no longer part
of the default installation.

Instead, users who would like to use SSH operations
should install ARIA's extra "[ssh]", which would install
Fabric and allow to take advantage of the execution-plugin's
SSH capabilities.

Users who won't install this extra will still be able to use
ARIA as well as the execution plugin, only without SSH.

Additional changes:

 - A new tox environment has been created for running
   SSH tests. The remaining envs only install plain ARIA.

 - requirements.in commented lines were removed -
   the bug that used to exist regarding environment markers
   has been fixed, and there's no longer the need
   to copy these manually to requirements.txt.

 - Environment-marked dependencies are now installed
   via "install_requires" rather than "extra_requires".

 - Added requirements.in to the manifest file,
   as well as fixed a bug in setup.py, which caused
   source distribution to make aria get installed
   without any dependencies before this fix.


asfgit:

PR #166, opened 2017-06-27
ARIA-294 workflow tasks execution is not in order

mxmrlv:


asfgit:

aria/modeling/orchestration.py
```diff
@@ -397,13 +397,8 @@ def retry(message=None, retry_interval=None):
...
 
     @declared_attr
-    def dependency_fk(self):
-        return relationship.foreign_key('task', nullable=True)
-
-    @declared_attr
     def dependencies(cls):
-        # symmetric relationship causes funky graphs
-        return relationship.one_to_many_self(cls, 'dependency_fk')
+        return relationship.many_to_many(cls, 'task', self=True)
```
mxmrlv: no need for other model if self=true

PR #167, opened 2017-06-28
ARIA-289 Don't attempt resuming a workflow which finished successfully

mxmrlv:


asfgit:

PR #169, opened 2017-06-29
ARIA-295 Makefile fixes

ran-z:
Fixed the "clean" target which would sometimes raise errors.
Removed the "deploy" target which was broken, and should probably
not be a part of the Makefile at this stage.


asfgit:

PR #170, opened 2017-06-30
ARIA-287 Add tox environment for docs

tliron:
Also adds the environment to "make test". Involves fixing Sphinx to
properly exclude SSH documentation when Fabric is not installed.


asfgit:

PR #174, opened 2017-07-04
ARIA-202 Execution plugin assumes /tmp for temp directory on the loc…

mxmrlv:
…al\/remote machine


asfgit:

PR #176, opened 2017-07-05
ARIA-237 Support for resuming failed workflow executions

mxmrlv:


aria/orchestrator/workflows/core/engine.py
```diff
@@ -68,8 +69,15 @@ def execute(self, ctx, resuming=False):
             if cancel:
                 self._terminate_tasks(tasks_tracker.executing_tasks)
                 events.on_cancelled_workflow_signal.send(ctx)
-            else:
+            elif all(task.status == task.SUCCESS or task.ignore_failure
```
ran-z: Is this really necessary?

aria/orchestrator/workflow_runner.py
```diff
@@ -38,7 +38,8 @@
 class WorkflowRunner(object):
 
     def __init__(self, model_storage, resource_storage, plugin_manager,
-                 execution_id=None, service_id=None, workflow_name=None, inputs=None, executor=None,
+                 execution_id=None, retry_failed=False,
```
ran-z: replace `retry_failed` with `retry_failed_tasks`

ran-z: expose `retry_failed` via CLI

tests/orchestrator/test_workflow_runner.py
```diff
@@ -474,10 +483,62 @@ def test_resume_failed_task(self, workflow_context, thread_executor):
...
+                execution_id=wf_runner.execution.id,
+                executor=new_thread_executor)
+
+            new_wf_runner.execute()
+        finally:
+            new_thread_executor.close()
+
+        # Wait for it to finish and assert changes.
+        node = workflow_context.model.node.refresh(node)
+        assert node.attributes['invocations'].value == 3
```
ran-z: Add an assert for validating the retry_count is reset back to 0 when using `retry_failed` = `True`

tests/orchestrator/test_workflow_runner.py
```diff
@@ -474,10 +483,62 @@ def test_resume_failed_task(self, workflow_context, thread_executor):
             new_thread_executor.close()
 
         # Wait for it to finish and assert changes.
+        node = workflow_context.model.node.refresh(node)
         assert node.attributes['invocations'].value == task.max_attempts - 1
         assert task.status == task.SUCCESS
         assert wf_runner.execution.status == wf_runner.execution.SUCCEEDED
 
+    def test_resume_failed_task_and_successful_task(self, workflow_context, thread_executor):
```
ran-z: Add one more test which runs a workflow (`retry_failed` = `False`) where the first task fails, the second one depends on the first one, and on resume validate that the first task doesn't run again but the second one does and succeeds.

mxmrlv: tests were added.

PR #178, opened 2017-07-06
ARIA-301 Fix issue when installing from wheel

ran-z:
Fixed an issue when installing ARIA from wheel where
environment-marked dependencies would install regardless
of the environment ARIA is installed on.


asfgit:

PR #180, opened 2017-07-10
ARIA-305 Advance ARIA version

ran-z:
Set package version to 0.2.0


asfgit:

PR #181, opened 2017-07-10
ARIA-103 Remove dependency on Clint

tliron:
We no longer require this third-party library, instead the utils/console
module uses the existing cli/color module.

This commit also fixes the cli/color module to properly support Unicode,
and also properly deinitialize Colorama in Windows.


asfgit:

PR #182, opened 2017-07-10
ARIA-312 Validation of workflow and operation kwargs raise False alarms

mxmrlv:
Workflow and Operation function kwargs validation failed in some scenarios.


asfgit:

PR #183, opened 2017-07-10
ARIA-76 Parallelize PyTest

tliron:
Use the PyTest xdist plugin to parallelize tests in boxed subprocesses.
Through benchmarking we discovered that paralellizing on the number of
CPU cores ("-n auto") provides the best times.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,23 +11,34 @@
 # limitations under the License.
 
 sudo: false
+
 language: python
+
 python:
   - "2.7"
+
 env:
-- TOX_ENV=pylint_code
-- TOX_ENV=pylint_tests
-- TOX_ENV=py27
-- TOX_ENV=py26
-- TOX_ENV=py27e2e
-- TOX_ENV=py26e2e
-- TOX_ENV=py27ssh
-- TOX_ENV=py26ssh
+  - TOX_ENV=pylint_code
+  - TOX_ENV=pylint_tests
+  - TOX_ENV=py27
+  - TOX_ENV=py26
+  - TOX_ENV=py27e2e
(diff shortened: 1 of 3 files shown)


asfgit:

asfgit:

.travis.yml
```diff
@@ -23,11 +23,12 @@ env:
...
+- TOX_ENV=docs
 install:
   - pip install --upgrade pip
   - pip install --upgrade setuptools
   - pip install tox
 script:
   - pip --version
   - tox --version
-  - tox -e $TOX_ENV
+  - PYTEST_PROCESSES=1 tox -e $TOX_ENV
```
mxmrlv: rename to PYTEST_CONCURRENCY

mxmrlv: What about appveyor? no concurrency?

mxmrlv: can we add some documentation, since it looks kinda awkward

mxmrlv: can't we just export this as environment variable?

tliron: No, there's is no script context here, it's one command at a time. :(

tliron: +1

tliron: Yes AppVeyor and yes concurrency. :) AppVeyor works fine with the `-n auto` defaults, likely because it's running in a restricted VM. The problem is only with Travis, because the OS reports a very large amount of CPUs (12 in our case) but the container is extremely restricted, and thus we get failures.

tliron: -1, because actually the name of the xdist argument is `--numprocesses`. Actually, it makes sense: the way xdist works is exactly by creating boxed subprocesses.

PR #187, opened 2017-07-26
ARIA-313 Fix handling the `required` field of inputs

AviaE:
The required field is handled in the following way:

topology_template inputs:
-------------------------
Every input that is declared as required must be supplied a value while
creating the service.
In addition, supplying inputs that were not declared in the topology
template is forbidden.

workflow inputs:*
----------------
Every input that is declared as required must be supplied a value while
creating/starting the execution.
In addition, supplying inputs that were not declared in the policy_type
is forbidden.
* workflow inputs are defined as properties of policy_types that are
derived from aria.Workflow

operation and interface inputs:
-------------------------------
The validation of the required field of inputs that belong to
operations and interfaces is done only in the parsing stage.
This reasoning follows the TOSCA spirit, where anything that is declared
as required in the type, must be assigned in the corresponding template

I split the logic of merging provided and declared input values into
three steps:

1. Validate that no undeclared inputs were provided.
2. Validate that all required inputs were provided with a value.
3. The actual merging process, which includes type checking.

### DIFF
--- a/extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
+++ b/extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
@@ -94,12 +94,10 @@ def create_service_template_model(context): # pylint: disable=too-many-locals,to
     # Topology template
     topology_template = context.presentation.get('service_template', 'topology_template')
     if topology_template is not None:
-        create_parameter_models_from_values(model.inputs,
-                                            topology_template._get_input_values(context),
-                                            model_cls=Input)
-        create_parameter_models_from_values(model.outputs,
-                                            topology_template._get_output_values(context),
-                                            model_cls=Output)
+        model.inputs.update(
+            create_input_models_from_values(topology_template._get_input_values(context)))
+        model.outputs.update(
+            create_output_models_from_values(topology_template._get_output_values(context)))
 
     # Plugin specifications
     policies = context.presentation.get('service_template', 'topology_template', 'policies')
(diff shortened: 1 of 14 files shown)


asfgit:

extensions/aria_extension_tosca/simple_v1_0/modeling/__init__.py
```diff
@@ -577,14 +554,44 @@ def added_all():
                         container.children.append(model)
 
 
-def create_parameter_models_from_values(properties, source_properties, model_cls):
+def create_input_models_from_values(inputs_field, template_inputs):
```
mxmrlv: Try and remove in-place update....

aria/orchestrator/workflows/api/task.py
```diff
@@ -137,6 +137,11 @@ def __init__(self,
         operation = self.actor.interfaces[self.interface_name].operations[self.operation_name]
         self.plugin = operation.plugin
         self.function = operation.function
+
+        modeling_utils.validate_required_inputs_are_supplied(
+            declared_inputs=operation.inputs,
+            supplied_inputs=operation.arguments)
```
AviaE: After consulting with @tliron and reviewing the TOSCA spec, I came to the conclusion that the validation of the `required` field of inputs that belong to operations and interfaces (as opposed to topology template and workflow inputs) should be done in the parsing stage.
This reasoning follows the TOSCA spirit, where anything that is declared in the type and is required, must be assigned in the corresponding template.

PR #188, opened 2017-07-28
ARIA-324 Refactor ctx proxy access

tliron:
Our previous use of "." to delimit nested dict keys was wrong (keys
could have a ".") and inflexible. The new implementation uses subsequent
args to move into the dict.

The same format can now be used to access object attributes.

This commit also changes how to support setting values: we must now use
"=" as the penultimate argument with the new value following.

Also fixed: callables will now "grab" the number of args they need
instead of all remaining args, making it possible to do further
inspection on the returned value from the callable. To allow for this,
kwargs are now expected as the first arg rather than the last.

Furthmore, this commit fixes a bad null check in the ctx client, and
also allows it to retrieve Unicode data.

### DIFF
--- a/aria/orchestrator/context/common.py
+++ b/aria/orchestrator/context/common.py
@@ -38,10 +38,27 @@ class BaseContext(object):
     """
 
     INSTRUMENTATION_FIELDS = (
+        modeling.models.Service.inputs,
+        modeling.models.ServiceTemplate.inputs,
+        modeling.models.Policy.properties,
+        modeling.models.PolicyTemplate.properties,
         modeling.models.Node.attributes,
         modeling.models.Node.properties,
         modeling.models.NodeTemplate.attributes,
-        modeling.models.NodeTemplate.properties
+        modeling.models.NodeTemplate.properties,
+        modeling.models.Group.properties,
+        modeling.models.GroupTemplate.properties,
+        modeling.models.Capability.properties,
+        # TODO ARIA-279: modeling.models.Capability.attributes,
+        modeling.models.CapabilityTemplate.properties,
+        # TODO ARIA-279: modeling.models.CapabilityTemplate.attributes
+        modeling.models.Relationship.properties,
+        modeling.models.Artifact.properties,
+        modeling.models.ArtifactTemplate.properties,
+        modeling.models.Interface.inputs,
+        modeling.models.InterfaceTemplate.inputs,
+        modeling.models.Operation.inputs,
+        modeling.models.OperationTemplate.inputs
     )
 
     class PrefixedLogger(object):
--- a/aria/orchestrator/execution_plugin/ctx_proxy/server.py
+++ b/aria/orchestrator/execution_plugin/ctx_proxy/server.py
@@ -17,14 +17,12 @@
 ``ctx`` proxy server implementation.
 """
 
-import collections
 import json
-import re
 import socket
-import threading
-import traceback
 import Queue
 import StringIO
+import threading
+import traceback
 import wsgiref.simple_server
 
 import bottle
@@ -123,9 +121,7 @@ class CtxProxy(object):
     def _process(self, request):
         try:
             with self.ctx.model.instrument(*self.ctx.INSTRUMENTATION_FIELDS):
-                typed_request = json.loads(request)
-                args = typed_request['args']
-                payload = _process_ctx_request(self.ctx, args)
+                payload = _process_request(self.ctx, request)
                 result_type = 'result'
                 if isinstance(payload, exceptions.ScriptException):
                     payload = dict(message=str(payload))
@@ -150,106 +146,94 @@ class CtxProxy(object):
         self.close()
 
 
-def _process_ctx_request(ctx, args):
-    current = ctx
-    num_args = len(args)
-    index = 0
-    while index < num_args:
-        arg = args[index]
-        attr = _desugar_attr(current, arg)
-        if attr:
-            current = getattr(current, attr)
-        elif isinstance(current, collections.MutableMapping):
-            key = arg
-            path_dict = _PathDictAccess(current)
-            if index + 1 == num_args:
-                # read dict prop by path
-                value = path_dict.get(key)
-                current = value
-            elif index + 2 == num_args:
-                # set dict prop by path
-                value = args[index + 1]
-                path_dict.set(key, value)
-                current = None
-            else:
-                raise RuntimeError('Illegal argument while accessing dict')
-            break
-        elif callable(current):
-            kwargs = {}
-            remaining_args = args[index:]
-            if isinstance(remaining_args[-1], collections.MutableMapping):
-                kwargs = remaining_args[-1]
-                remaining_args = remaining_args[:-1]
-            current = current(*remaining_args, **kwargs)
-            break
+class CtxError(RuntimeError):
+    pass
+
+
+class CtxParsingError(CtxError):
+    pass
+
+
+def _process_request(ctx, request):
+    request = json.loads(request)
+    args = request['args']
+    return _process_arguments(ctx, args)
+
+
+def _process_arguments(obj, args):
+    # Modifying?
+    try:
+        # TODO: should there be a way to escape "=" in case it is needed as real argument?
+        equals_index = args.index('=') # raises ValueError if not found
+    except ValueError:
+        equals_index = None
+    if equals_index is not None:
+        if equals_index == 0:
+            raise CtxParsingError('The "=" argument cannot be first')
+        elif equals_index != len(args) - 2:
+            raise CtxParsingError('The "=" argument must be penultimate')
+        modifying = True
+        modifying_key = args[-3]
+        modifying_value = args[-1]
+        args = args[:-3]
+    else:
+        modifying = False
+        modifying_key = None
+        modifying_value = None
+
+    # Parse all arguments
+    while len(args) > 0:
+        obj, args = _process_next_operation(obj, args, modifying)
+
+    if modifying:
+        if hasattr(obj, '__setitem__'):
+            # Modify item value (dict, list, and similar)
+            if isinstance(obj, (list, tuple)):
+                modifying_key = int(modifying_key)
+            obj[modifying_key] = modifying_value
+        elif hasattr(obj, modifying_key):
+            # Modify object attribute
+            setattr(obj, modifying_key, modifying_value)
         else:
-            raise RuntimeError('{0} cannot be processed in {1}'.format(arg, args))
-        index += 1
-    if callable(current):
-        current = current()
-    return current
+            raise CtxError('Cannot modify `{0}` of `{1!r}`'.format(modifying_key, obj))
+
+    return obj
 
 
-def _desugar_attr(obj, attr):
-    if not isinstance(attr, basestring):
-        return None
-    if hasattr(obj, attr):
-        return attr
-    attr = attr.replace('-', '_')
-    if hasattr(obj, attr):
-        return attr
-    return None
+def _process_next_operation(obj, args, modifying):
+    args = list(args)
+    arg = args.pop(0)
 
+    # Call?
+    if arg == '[':
+        # TODO: should there be a way to escape "[" and "]" in case they are needed as real
+        # arguments?
+        try:
+            closing_index = args.index(']') # raises ValueError if not found
+        except ValueError:
+            raise CtxParsingError('Opening "[" without a closing "]')
+        callable_args = args[:closing_index]
+        args = args[closing_index + 1:]
+        if not callable(obj):
+            raise CtxError('Used "[" and "] on an object that is not callable')
+        return obj(*callable_args), args
 
-class _PathDictAccess(object):
-    pattern = re.compile(r"(.+)\[(\d+)\]")
+    # Attribute?
+    if isinstance(arg, basestring):
+        if hasattr(obj, arg):
+            return getattr(obj, arg), args
+        token_sugared = arg.replace('-', '_')
+        if hasattr(obj, token_sugared):
+            return getattr(obj, token_sugared), args
 
-    def __init__(self, obj):
-        self.obj = obj
+    # Item? (dict, lists, and similar)
+    if hasattr(obj, '__getitem__'):
+        if modifying and (arg not in obj) and hasattr(obj, '__setitem__'):
+            # Create nested dict
+            obj[arg] = {}
+        return obj[arg], args
 
-    def set(self, prop_path, value):
-        obj, prop_name = self._get_parent_obj_prop_name_by_path(prop_path)
-        obj[prop_name] = value
-
-    def get(self, prop_path):
-        value = self._get_object_by_path(prop_path)
-        return value
-
-    def _get_object_by_path(self, prop_path, fail_on_missing=True):
-        # when setting a nested object, make sure to also set all the
-        # intermediate path objects
-        current = self.obj
-        for prop_segment in prop_path.split('.'):
-            match = self.pattern.match(prop_segment)
-            if match:
-                index = int(match.group(2))
-                property_name = match.group(1)
-                if property_name not in current:
-                    self._raise_illegal(prop_path)
-                if not isinstance(current[property_name], list):
-                    self._raise_illegal(prop_path)
-                current = current[property_name][index]
-            else:
-                if prop_segment not in current:
-                    if fail_on_missing:
-                        self._raise_illegal(prop_path)
-                    else:
-                        current[prop_segment] = {}
-                current = current[prop_segment]
-        return current
-
-    def _get_parent_obj_prop_name_by_path(self, prop_path):
-        split = prop_path.split('.')
-        if len(split) == 1:
-            return self.obj, prop_path
-        parent_path = '.'.join(split[:-1])
-        parent_obj = self._get_object_by_path(parent_path, fail_on_missing=False)
-        prop_name = split[-1]
-        return parent_obj, prop_name
-
-    @staticmethod
-    def _raise_illegal(prop_path):
-        raise RuntimeError('illegal path: {0}'.format(prop_path))
+    raise CtxParsingError('Cannot parse argument: `{0!r}`'.format(arg))
 
 
 def _get_unused_port():
--- a/aria/orchestrator/workflows/api/task.py
+++ b/aria/orchestrator/workflows/api/task.py
@@ -140,13 +140,18 @@ class OperationTask(BaseTask):
         self.arguments = modeling_utils.merge_parameter_values(arguments,
                                                                operation.arguments,
                                                                model_cls=models.Argument)
-        if getattr(self.actor, 'outbound_relationships', None) is not None:
+
+        actor = self.actor
+        if hasattr(actor, '_wrapped'):
+            # Unwrap instrumented model
+            actor = actor._wrapped
+        if isinstance(actor, models.Node):
             self._context_cls = context.operation.NodeOperationContext
-        elif getattr(self.actor, 'source_node', None) is not None:
+        elif isinstance(actor, models.Relationship):
             self._context_cls = context.operation.RelationshipOperationContext
         else:
-            raise exceptions.TaskCreationException('Could not locate valid context for '
-                                                   '{actor.__class__}'.format(actor=self.actor))
+            raise exceptions.TaskCreationException('Could not create valid context for '
+                                                   '{actor.__class__}'.format(actor=actor))
 
     def __repr__(self):
         return self.name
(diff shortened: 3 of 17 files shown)


asfgit:

aria/orchestrator/execution_plugin/ctx_proxy/server.py
```diff
@@ -150,43 +150,101 @@ def __exit__(self, *args, **kwargs):
...
+                spec_args = argspec.args[len(current.args):]
+                # Don't count keyword args fulfilled by the partial
+                spec_args = tuple(a for a in spec_args if a not in current.keywords)
+            else:
+                argspec = inspect.getargspec(current)
+                spec_args = argspec.args
+
+            callable_kwargs = {}
+            if isinstance(arg, dict):
+                # If the first arg is a dict, treat it as our kwargs
```
mxmrlv: Indeed, this leads to an odd behavior

aria/orchestrator/execution_plugin/ctx_proxy/server.py
```diff
@@ -150,43 +150,101 @@ def __exit__(self, *args, **kwargs):
...
+        # Call
         elif callable(current):
-            kwargs = {}
-            remaining_args = args[index:]
-            if isinstance(remaining_args[-1], collections.MutableMapping):
-                kwargs = remaining_args[-1]
-                remaining_args = remaining_args[:-1]
-            current = current(*remaining_args, **kwargs)
-            break
+            if isinstance(current, functools.partial):
```
mxmrlv: Maybe add some documentation why we inspect the called method

aria/orchestrator/execution_plugin/ctx_proxy/server.py
```diff
@@ -150,43 +150,101 @@ def __exit__(self, *args, **kwargs):
...
+                # Gobble the rest of the args
+                callable_args = args[index:]
+            else:
+                # Take only what we need
+                spec_args = tuple(a for a in spec_args if a not in callable_kwargs)
+                spec_args_count = len(spec_args)
+                if inspect.ismethod(current):
+                    # Don't count "self" argument
+                    spec_args_count -= 1
+                callable_args = args[index:index + spec_args_count]
```
mxmrlv: what about the case you have extra args, these will be silently ignored.
Why can't we just "spill" the remaining args/kwargs into the function?

aria/orchestrator/execution_plugin/ctx_proxy/server.py
```diff
@@ -150,43 +150,101 @@ def __exit__(self, *args, **kwargs):
...
-                raise RuntimeError('Illegal argument while accessing dict')
-            break
+
+        # List entry
+        elif isinstance(current, list) and _is_int(arg):
+            current = current[int(arg)]
+
+        # Dict (or dict-like object) value
+        elif hasattr(current, '__getitem__'):
+            if modifying and (not arg in current):
```
mxmrlv: According to these semantics we lose any knowledge of whether we are accessing a list or a dict

aria/orchestrator/execution_plugin/ctx_proxy/server.py
```diff
@@ -150,43 +150,101 @@ def __exit__(self, *args, **kwargs):
...
 
-def _process_ctx_request(ctx, args):
+def _process_ctx_request(ctx, args): # pylint: disable=too-many-branches,too-many-statements
     current = ctx
-    num_args = len(args)
     index = 0
+
+    try:
+        # TODO: should there be a way to escape "=" in case it is needed as real argument?
+        equals_index = args.index('=')
```
AviaE: Maybe add a comment here that says that `.index` raises a `ValueError` if the value is not in the list. And that is why you check for `ValueError` in the end. It wasn't clear for @mxmrlv also.
In addition, why does all the other logic under this line is in the `try` if it doesn't expected to raise `ValueError`. I understand that extracting this part will force you to refactor the order of the statements a bit - just a thought =)

aria/orchestrator/execution_plugin/ctx_proxy/server.py
```diff
@@ -150,43 +150,101 @@ def __exit__(self, *args, **kwargs):
...
+
         index += 1
+
     if callable(current):
         current = current()
+
+    if modifying:
+        if hasattr(current, '__setitem__'):
+            # Modify dict value
+            current[modifying_key] = modifying_value
```
AviaE: since `modifying_value` is a string, this raises an error when trying to modify the value of a list element.
So in the current implementation you won't able to do thing of the sorts:
`node properties prop5 0 = new_value`
(I tested this)

tests/orchestrator/execution_plugin/test_ctx_proxy_server.py
```diff
@@ -39,26 +39,26 @@ def test_dict_prop_access_get_key(self, server):
...
 
     def test_dict_prop_access_get_key_nested(self, server):
-        response = self.request(server, 'node', 'properties', 'prop2.nested_prop1')
+        response = self.request(server, 'node', 'properties', 'prop2', 'nested_prop1')
         assert response == 'nested_value1'
 
     def test_dict_prop_access_get_with_list_index(self, server):
-        response = self.request(server, 'node', 'properties', 'prop3[2].value')
+        response = self.request(server, 'node', 'properties', 'prop3', 2, 'value')
         assert response == 'value_2'
```
AviaE: Add a test for modifying a list element.

aria/orchestrator/context/common.py
```diff
@@ -38,10 +38,27 @@ class BaseContext(object):
     """
 
     INSTRUMENTATION_FIELDS = (
+        modeling.models.Service.inputs,
```
tliron: Because you should be able to do things like: {{ctx node capabilities host properties username}}

The ctx proxy should let you access all models.

aria/orchestrator/workflows/api/task.py
```diff
@@ -140,13 +140,18 @@ def __init__(self,
         self.arguments = modeling_utils.merge_parameter_values(arguments,
                                                                operation.arguments,
                                                                model_cls=models.Argument)
-        if getattr(self.actor, 'outbound_relationships', None) is not None:
```
tliron: Why not import models? This should have been explained in comments. Duck typing seems very inaccurate and prone to errors.

aria/orchestrator/workflows/api/task.py
```diff
@@ -140,13 +140,18 @@ def __init__(self,
...
+
+        actor = self.actor
+        if hasattr(actor, '_wrapped'):
```
tliron: Because you often get instrumented/wrapped models here, and they would fail the `isinstance` check. (With duck typing check they succeed, but I thought duck typing was bad.)

aria/storage/collection_instrumentation.py
```diff
@@ -213,32 +235,32 @@ def __init__(self, mapi, *args, **kwargs):
...
-        return_value = getattr(self._wrapped, item)
-        if isinstance(return_value, self._wrapped.__class__):
-            return _create_instrumented_model(return_value, self._mapi, self._instrumentation)
-        if isinstance(return_value, (list, dict)):
-            return _create_wrapped_model(return_value, self._mapi, self._instrumentation)
-        return return_value
-
     def _apply_instrumentation(self):
         for field in self._instrumentation:
+            if field.parent.class_ != type(self._wrapped):
```
tliron: It doesn't seem to break this in tests.

Without this check, the only check is by field name, which caused a lot of breakage. You would node properties being instrumented for capability properties. Again, consider the use case of `ctx node capabilities host properties username`.

aria/orchestrator/execution_plugin/ctx_proxy/server.py
```diff
@@ -150,43 +150,101 @@ def __exit__(self, *args, **kwargs):
         self.close()
 
 
-def _process_ctx_request(ctx, args):
+def _process_ctx_request(ctx, args): # pylint: disable=too-many-branches,too-many-statements
```
tliron: Of course we can, but it doesn't seem a very long function at all. This is one of those cases where PyLint is not being helpful.

aria/orchestrator/execution_plugin/ctx_proxy/server.py
```diff
@@ -150,43 +150,101 @@ def __exit__(self, *args, **kwargs):
...
     current = ctx
-    num_args = len(args)
     index = 0
+
```
tliron: This was the whole point of this PR.

The previous implementation assumed either one arg or two: one arg meant "read" and two args meant "write". But this required us to support dot notation so that the one arg could refer to a deep nested attribute. Removing dot notation requires a different to support "write".

This PR also supports working on the return value of function calls.

aria/orchestrator/execution_plugin/ctx_proxy/server.py
```diff
@@ -150,43 +150,101 @@ def __exit__(self, *args, **kwargs):
...
-                current = None
-            else:
-                raise RuntimeError('Illegal argument while accessing dict')
-            break
+
+        # List entry
+        elif isinstance(current, list) and _is_int(arg):
+            current = current[int(arg)]
+
+        # Dict (or dict-like object) value
```
tliron: Because that is broken on wrapped instrumented models...

aria/orchestrator/execution_plugin/ctx_proxy/server.py
tliron: +1

tliron: I don't understand this comment.

aria/orchestrator/execution_plugin/ctx_proxy/server.py
tliron: This happens a lot, actually. I forget which of our tests would showcase this exactly, but I think even `ctx logger` is based on partials.

tliron: +1

The main reason for all of this is that we sometimes need to continue access properties on returned values on callables. The previous implementation simpled "gobbled" all remaining args, but we might need those args. Example from Clearwater:

`ctx service get_policy_by_type clearwater.Configuration properties zone`

But generally I suggest you take a look at the various uses of `ctx` in the [Clearwater example](https://github.com/apache/incubator-ariatosca/blob/ARIA-321-clearwater/examples/clearwater/scripts/host/configure.sh).

aria/orchestrator/execution_plugin/ctx_proxy/server.py
tliron: See comment above: we can't just grab all remaining args, because we might need these args for further delving into the returned value of the callable.

aria/orchestrator/execution_plugin/ctx_proxy/server.py
tliron: +1

aria/orchestrator/execution_plugin/ctx_proxy/server.py
tliron: +1

Will add a test for this, too.

tests/orchestrator/execution_plugin/test_ctx_proxy_server.py
tliron: +1

tliron:
Hey guys, I suggest going over the complex use cases in the [Clearwater example](https://github.com/apache/incubator-ariatosca/blob/ARIA-321-clearwater/examples/clearwater/scripts/host/configure.sh).

Basically all the changes in the PR were inspired by those real-world needs of `ctx`.

aria/storage/collection_instrumentation.py
tliron: To fix: check by isubclass instead of equality

aria/orchestrator/execution_plugin/ctx_proxy/server.py
```diff
@@ -170,101 +178,59 @@ def _process_ctx_request(ctx, args): # pylint: disable=too-many-branches,too-man
...
+        if hasattr(obj, arg):
+            return getattr(obj, arg), args
+        arg_sugared = arg.replace('-', '_')
+        if hasattr(obj, arg_sugared):
+            return getattr(obj, arg_sugared), args
+
+    # Item? (dict, lists, and similar)
+    if hasattr(obj, '__getitem__'):
+        if modifying and hasattr(obj, 'has_key') and (not obj.has_key(arg)):
+            obj[arg] = {}
```
mxmrlv: why is it set to a dict? couldn't it be a list?

aria/orchestrator/execution_plugin/ctx_proxy/server.py
```diff
@@ -170,101 +178,59 @@ def _process_ctx_request(ctx, args): # pylint: disable=too-many-branches,too-man
...
-    if not isinstance(attr, basestring):
-        return None
-    if hasattr(obj, attr):
-        return attr
-    attr = attr.replace('-', '_')
-    if hasattr(obj, attr):
-        return attr
-    return None
 
+def _parse_argument(obj, args, modifying):
```
mxmrlv: consider renaming it, since it accepts all the args (and not a single arg)

mxmrlv: i would consider splitting this function into two different functions (since the only usage for the entire args is being made in case of a function call

aria/orchestrator/execution_plugin/ctx_proxy/server.py
```diff
@@ -170,101 +178,59 @@ def _process_ctx_request(ctx, args): # pylint: disable=too-many-branches,too-man
...
+    # Attribute?
+    if isinstance(arg, basestring):
+        if hasattr(obj, arg):
+            return getattr(obj, arg), args
+        arg_sugared = arg.replace('-', '_')
+        if hasattr(obj, arg_sugared):
+            return getattr(obj, arg_sugared), args
+
+    # Item? (dict, lists, and similar)
+    if hasattr(obj, '__getitem__'):
```
mxmrlv: consider `isinstance(obj, (list, dict))`

aria/orchestrator/execution_plugin/ctx_proxy/server.py
```diff
@@ -170,101 +178,59 @@ def _process_ctx_request(ctx, args): # pylint: disable=too-many-branches,too-man
...
+    # Parse all arguments
+    while len(args) > 0:
+        obj, args = _parse_argument(obj, args, modifying)
 
     if modifying:
-        if hasattr(current, '__setitem__'):
-            # Modify dict value
-            current[modifying_key] = modifying_value
-        else:
+        if hasattr(obj, '__setitem__'):
```
mxmrlv: consider `isinstance(obj, (dict, list))`

aria/orchestrator/execution_plugin/ctx_proxy/server.py
```diff
@@ -150,17 +146,29 @@ def __exit__(self, *args, **kwargs):
         self.close()
 
 
-def _process_ctx_request(ctx, args): # pylint: disable=too-many-branches,too-many-statements
-    current = ctx
-    index = 0
+class ProcessingError(RuntimeError):
```
mxmrlv: consider `CtxArgProcessingError`, or any other name since `ProcessingError` makes it sound like its a process related issue.

examples/hello-world/scripts/stop.sh
```diff
@@ -16,14 +16,13 @@
...
-PID=`cat ${PYTHON_FILE_SERVER_ROOT}/${PID_FILE}`
+ctx logger info [ "Shutting down web server, pid = ${PID}." ]
+kill -9 "$PID" || exit $?
 
-ctx logger info "Shutting down web server, pid = ${PID}."
-kill -9 ${PID} || exit $?
-
-ctx logger info "Removing web server root folder: ${PYTHON_FILE_SERVER_ROOT}."
-rm -rf ${PYTHON_FILE_SERVER_ROOT}
+ctx logger info [ "Removing web server root folder: $PYTHON_FILE_SERVER_ROOT." ]
```
tliron: Heh, not 100% sure I agree with that, but it does look clear. :)

tests/orchestrator/execution_plugin/test_ctx_proxy_server.py
```diff
@@ -55,61 +55,32 @@ def test_dict_prop_access_set(self, server, ctx):
         assert ctx.node.properties['prop3'][2]['value'] == 'new_value_2'
         assert ctx.node.properties['prop4']['some']['new']['path'] == 'some_new_value'
 
+    def test_dict_prop_access_set_with_list_index(self, server, ctx):
```
tliron: +1

aria/orchestrator/execution_plugin/ctx_proxy/server.py
tliron: Cannot, because it wouldn't support dict-like objects like instrumented wrapper classes. This is generally better because it would support anything that supported __get_item__ and __set_item__, which allows for any kind of classes.

aria/orchestrator/execution_plugin/ctx_proxy/server.py
```diff
@@ -150,17 +146,29 @@ def __exit__(self, *args, **kwargs):
...
+    args = request['args']
+    return _parse_arguments(ctx, args)
+
+
+def _parse_arguments(obj, args):
+    # Modifying?
     try:
         # TODO: should there be a way to escape "=" in case it is needed as real argument?
-        equals_index = args.index('=')
+        equals_index = args.index('=') # raises ValueError if not found
```
tliron: It looks more awkward, but +1.

aria/orchestrator/execution_plugin/ctx_proxy/server.py
tliron: But you wouldn't know in advance which function to call until you start parsing. I renamed it and did some little cleanups to add clarity.

aria/orchestrator/execution_plugin/ctx_proxy/server.py
```diff
@@ -170,101 +178,59 @@ def _process_ctx_request(ctx, args): # pylint: disable=too-many-branches,too-man
...
+    if isinstance(arg, basestring):
+        if hasattr(obj, arg):
+            return getattr(obj, arg), args
+        arg_sugared = arg.replace('-', '_')
+        if hasattr(obj, arg_sugared):
+            return getattr(obj, arg_sugared), args
+
+    # Item? (dict, lists, and similar)
+    if hasattr(obj, '__getitem__'):
+        if modifying and hasattr(obj, 'has_key') and (not obj.has_key(arg)):
```
tliron: Not a good idea. :( Because the "not in" requires an implementation of __iter__, but __iter__ might not necessarily adhere to the dict contract of iterating only keys. (I did add the "right" kind of __iter__ to the instrumented wrapper, but nobody can promise us that other objects would behave as expected here.

aria/orchestrator/execution_plugin/ctx_proxy/server.py
tliron: I think it's non-trivial and not useful to set nested lists that have no value. Consider this:
```
x = ['a','b','c']
ctx x 10 name = 2
```
What is to be done with indexes between 3 and 9 of the list? Fill them with None? With empty dicts? It is unclear what would be best. I think it is best in such situations for ARIA to fail with an error message. Only let users modify existing index values.

I recommend leaving it as is.

aria/orchestrator/execution_plugin/ctx_proxy/server.py
mxmrlv: ducktyping much? :smiley: but ok

aria/orchestrator/execution_plugin/ctx_proxy/server.py
mxmrlv: but who promises my the support of `has_key`?

aria/orchestrator/execution_plugin/ctx_proxy/server.py
mxmrlv: but what about ctx x 2 name = value?

aria/orchestrator/execution_plugin/ctx_proxy/server.py
tliron: Correct, this is *exactly* a case for ducktyping. :) Any object can be "dict-like" here by just implementing this magic method, and ctx will be happy.

aria/orchestrator/execution_plugin/ctx_proxy/server.py
tliron: That works, as long as index 2 already exists in `x`.

asfgit:

PR #190, opened 2017-08-15
ARIA-349-get_attribute-is-not-calculated-at-runtime

mxmrlv:


asfgit:

PR #194, opened 2017-08-30
ARIA-353 Add node template directives

AviaE:
The directives field already existed in the presentation level, but it
needed to be added to the node template model.


asfgit:

PR #198, opened 2017-10-23
ARIA-391 Increase wagon and setuptools dependency version

mxmrlv:


asfgit:

PR #199, opened 2017-10-23
ARIA-392 Failing to load ruamel.yaml

mxmrlv:


asfgit:

aria/__init__.py
```diff
@@ -17,14 +17,45 @@
...
 import pkg_resources
-
 aria_package_name = 'apache-ariatosca'
 __version__ = pkg_resources.get_distribution(aria_package_name).version
 
+
+
+
+try:
+    import ruamel                                           # noqa: F401
```
tliron: I think we've decided to remove IDE-specific comments.

aria/__init__.py
```diff
@@ -17,14 +17,45 @@
...
+except ImportError:
```
tliron: I recent commit removed Py2.6 support and allowed us to upgrade ruamel to its latest version. So, I suggest rebasing on master and testing again: perhaps this hack isn't needed anymore.

Assuming this hack is still needed, we need to find a way to generalize it, since we import ruamel in many, many places in the code. Perhaps we should have an {{aria.utils.yaml}} package where we do all this work once. Everywhere else we use yaml would then have to be replaced to {{from aria.utils.yaml import yaml}}. A nice side benefit would be that it would be easy to replace ruamel with a different yaml parser is that is ever necessary.

tliron: Another follow up: if this is still an issue with newer versions of ruaeml, I strongly suggest we open a bug at that project. I've done it in the past and have gotten things fixed. If you do that, make sure to link that bug report in a code comment here -- in case it's solved in the future, we should remove our workaround.

mxmrlv: The problem with ruamel is that it's missing the top level __init__. I have seen that the author of ruamel had no intention of fixing it. instead he supplied .pth files as part of the installation. This files do work in most cases but not when the destination dir is a custom dir. Placing this code under the `__init__.py` enables loading ruamel from each submodule of ARIA. I'm not sure how placing it under `aria.utils.yaml` would be a better fix. But if you feel it's better we could do that.

tliron: I just think `from aria.utils.yaml import yaml` will give us a clean separation in case we have a better fix in the future, so I do prefer that.

Also, for this PR to merge we need to change *all* our uses of ruamel in the codebase to that import.

mxmrlv:
@tliron any additional comments? or should i merge?

tliron:
@mxmrlv You didn't address these:

1. Let's move this to `from aria.utils.yaml import yaml`. I don't think the aria `__init__.py` is the right place for this.
2. Let's go over *all* places in the codebase where we import ruamel to use the above.

mxmrlv:
This code only runs once, it doesn't provide with a patched up ruamel, it patched the loading mechanism...so loading should work everywhere, as long as the code was run...

tliron:
Still, I don't see why it should be in `aria/__init__.py` and be always
activated. I think it should be in activated only for a module that
specifically is using ruamel. It would still only happen once.

mxmrlv:
OK, np just wanted to make sure we're on the same page...

tliron:
I think we are. Also: I only suggest, you of course can decide differently.
Maybe do one more commit to make sure we are on the same page? Also feel
free to rebase and squash, it's actually easier for a small PR like this.

mxmrlv:
so, i think we're ready

tliron:
OK. So I guess you don't want to change the imports in `/tests/`?

PR #206, opened 2017-10-30
ARIA-405 Remove support for Python 2.6

tliron:
* setup.py now requires Python 2.7
* Upgrade all 3rd party libraries to recent versions
* API changes to networkx
* Stricter yaml.load call for ruamel.yaml
* Remove iter_modules implementation for Python 2.6
* Remove NullHander implementation for Python 2.6
* Remove "py26" tox test environments


asfgit:

PR #209, opened 2017-11-29
ARIA-414 Current events handler mechanism relies on sqlalchemy

mxmrlv:


asfgit:

tests/orchestrator/workflows/executor/test_executor.py
```diff
@@ -28,45 +28,48 @@
...
     process,
     # celery
 )
 
 import tests
+from .. import helpers
 from . import MockContext
 
 
+
```
aviyoop: remove (if this is not the only change)

asfgit:

PR #210, opened 2017-12-03
ARIA-416 Providing handlers to logging_handlers only adds handles inst…

mxmrlv:
…ead of replacing

### DIFF
--- a/aria/orchestrator/context/common.py
+++ b/aria/orchestrator/context/common.py
@@ -115,14 +115,19 @@ class BaseContext(object):
 
     @contextmanager
     def logging_handlers(self, handlers=None):
+        original_handlers = self.logger.handlers
         handlers = handlers or []
         try:
             for handler in handlers:
                 self.logger.addHandler(handler)
+            for handler in original_handlers:
+                self.logger.removeHandler(handler)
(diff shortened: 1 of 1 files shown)


asfgit:

aria/orchestrator/context/common.py
```diff
@@ -115,14 +115,19 @@ def __repr__(self):
 
     @contextmanager
     def logging_handlers(self, handlers=None):
```
aviyoop: why would you call this context manager without providing handlers?

aria/orchestrator/context/common.py
```diff
@@ -115,14 +115,19 @@ def __repr__(self):
...
+        original_handlers = self.logger.handlers
         handlers = handlers or []
         try:
             for handler in handlers:
                 self.logger.addHandler(handler)
+            for handler in original_handlers:
+                self.logger.removeHandler(handler)
             yield self.logger
         finally:
             for handler in handlers:
```
aviyoop: To prevent the original handlers from being duplicated, you should first (in the `finally` block) empty the logger from handlers, and then add the original handlers.

PR #212, opened 2017-12-04

djay87:
This commit supports using the plugin.yaml present in a plugin wagon archive in a more simpler way as described below.

imports:
  - file: <plugin_name>-<plugin_version>
    repository: plugins


asfgit:

PR #213, opened 2017-12-04
ARIA-399: Sample Dockerfile and docker-compose.yaml file that installs/runs ARIA

vishwanathj:


asfgit:

tliron:
Thank you for this contribution. In my opinion it would be better to share these files on the wiki. This is not a TOSCA example, so it definitely doesn't belong in our examples dir. And I think generally it's more about how to package/deploy ARIA, which should be part of the instructions on the wiki.

vishwanathj:
Tal, 
Appreciate your valuable feedback.
Would creating a docker-examples similar to tosca-examples at https://cwiki.apache.org/confluence/display/ARIATOSCA/TOSCA+Examples work?

PR #220, opened 2017-12-07
ARIA-433 add missing enum column name

dusking:


asfgit:

PR #224, opened 2018-02-02
Add service proxy plugin example

dfilppi:
Plugin that implements a "service proxy".  Eases the composition of separate ARIA managed services by gather the outputs of services into the service proxy attributes.


asfgit:

dfilppi:
Need to fix build error

PR #226, opened 2018-02-12

djay87:
Support for importing plugin.yaml files of loaded plugins.


asfgit:

tests/extensions/aria_extension_tosca/simple_v1_0/test_imports.py
```diff
@@ -146,6 +149,21 @@ def test_import_repository(parser, repository):
       type: MyNode
 """, dict(repository=repository)).assert_success()
 
+#Plugin
```
tnadeauaria: Need a function header comment describing what this does, what goes in and what comes out.

PR #227, opened 2018-02-12

djay87:
Import "plugin.yaml" present in plugin resource storage via 'plugins' repository.


asfgit:

PR #230, opened 2018-05-16
ARIA-440 Service template creation fails while providing CSAR as input.

djay87:


asfgit:

